PR #1, opened 2016-04-16
add fb_apt cookbook

davide125:
This adds a cookbook to manage APT using the attribute-driven API. It's a rework of an unpublished cookbook I've been running at home for a while. It tries to fully manage everything around APT, including keys and configurations. Tested only on Debian but includes code to support Ubuntu as well; should be trivial to port to other Debian-based distros.

### DIFF
--- a/fb_apt/README.md
+++ b/fb_apt/README.md
@@ -23,11 +23,12 @@ Attributes
 
 Usage
 -----
-To install and configure APT include `fb_apt`, which will populate the 
-repository sources in `/etc/apt/sources.list` and update the package cache 
-during the run if it's old than `node['fb_apt']['update_delay']` seconds
-(defaults to 86400). This uses the `execute[apt-get update]` resource, which
-other cookbooks can suscribe to or notify as well.
+To install and configure APT include `fb_apt`, which will populate the
+repository sources in `/etc/apt/sources.list` and update the package cache
+during the run if it's older than `node['fb_apt']['update_delay']` seconds
+(defaults to 86400). To force an update on every Chef run, set this attribute
+to 0. The actual update is done via the `execute[apt-get update]` resource,
+which other cookbooks can suscribe to or notify as well.
 
 ### Repository sources
 By default the cookbook will setup the base distribution repos based on the
--- a/fb_apt/README.md
+++ b/fb_apt/README.md
@@ -25,9 +25,9 @@ Usage
 -----
 To install and configure APT include `fb_apt`, which will populate the 
 repository sources in `/etc/apt/sources.list` and update the package cache 
-every `node['fb_apt']['update_delay']` seconds (defaults to 86400) using the 
-`execute[apt-get update]` resource (which other cookbooks can suscribe to or 
-notify).
+during the run if it's old than `node['fb_apt']['update_delay']` seconds
+(defaults to 86400). This uses the `execute[apt-get update]` resource, which
+other cookbooks can suscribe to or notify as well.
 
 ### Repository sources
 By default the cookbook will setup the base distribution repos based on the
@@ -72,10 +72,6 @@ anything in `/etc/apt/apt.conf.d`. Example:
         'Proxy' => 'http://myproxy:3412',
       },
     })
-
-Internally, `fb_apt` uses the `FB::Apt.gen_apt_conf_entry()` library function
-to generate this config file. This is an internal function that's not meant to 
-be used by other cookbooks.
 
 ### Preferences
 You can fine tune which versions of packages will be selected for installation
--- a/fb_apt/README.md
+++ b/fb_apt/README.md
@@ -27,9 +27,7 @@ To install and configure APT include `fb_apt`, which will populate the
 repository sources in `/etc/apt/sources.list` and update the package cache 
 every `node['fb_apt']['update_delay']` seconds (defaults to 86400) using the 
 `execute[apt-get update]` resource (which other cookbooks can suscribe to or 
-notify against). Note that the `node['fb_apt']['update_delay']` attribute is 
-evaluated at compile time, so you'd need to set it before including `fb_apt` if
-you'd like to customize it.
+notify).
 
 ### Repository sources
 By default the cookbook will setup the base distribution repos based on the
@@ -52,13 +50,13 @@ Repository keys can be added to `node['fb_apt']['keys']` which is a hash in the
 from the `node['fb_apt']['keyserver']` keyserver (`keys.gnupg.net` by default).
 Example:
 
-  node.default['fb_apt']['keys'] = {
-    '94558F59' => nil,
-    'F3EFDBD9' => <<-eos
-  -----BEGIN PGP PUBLIC KEY BLOCK-----
-  ...
-  eos
-  }
+    node.default['fb_apt']['keys'] = {
+      '94558F59' => nil,
+      'F3EFDBD9' => <<-eos
+    -----BEGIN PGP PUBLIC KEY BLOCK-----
+    ...
+    eos
+    }
 
 Automatic key fetching can be disabled by setting the keyserver to `nil`; this 
 will produce an exception for any unspecified key. By default `fb_apt` will 
@@ -67,22 +65,25 @@ manage the keyring at `/etc/apt/trusted.gpg`; this can be customized with
 
 ### Configuration
 APT behaviour can be customized using `node['fb_apt']['config']`, which will be
-used to populate `/etc/apt/apt.conf`. Example:
+used to populate `/etc/apt/apt.conf`. Note that this will take precedence over
+anything in /etc/apt/apt.conf.d. Example:
 
-  node.default['fb_apt']['config'] = {
-    'Acquire::http' => {
+    node.default['fb_apt']['config'].merge!({
+      'Acquire::http' => {
       'Proxy' => 'http://myproxy:3412',
-    },
-  }
+      },
+    }
   
 ### Preferences
 You can fine tune which versions of packages will be selected for installation
-by tweaking APT preferences via `node['fb_apt']['preferences']`. Example:
+by tweaking APT preferences via `node['fb_apt']['preferences']`. Note that we
+clobber the contents of `/etc/apt/preferences.d` to ensure this always takes
+precedence. Example:
 
-  node.default['fb_apt']['preferences'] = {
-    'Pin dpatch package from experimental' => {
-      'Package' => 'dpatch',
-      'Pin' => 'release o=Debian,a=experimental',
-      'Pin-Priority' => 450,
+    node.default['fb_apt']['preferences'].merge!{
+      'Pin dpatch package from experimental' => {
+        'Package' => 'dpatch',
+        'Pin' => 'release o=Debian,a=experimental',
+        'Pin-Priority' => 450,
+      }
     }
-  }
--- a/fb_apt/README.md
+++ b/fb_apt/README.md
@@ -51,13 +51,11 @@ Repository keys can be added to `node['fb_apt']['keys']` which is a hash in the
 from the `node['fb_apt']['keyserver']` keyserver (`keys.gnupg.net` by default).
 Example:
 
-    node.default['fb_apt']['keys'] = {
-      '94558F59' => nil,
-      'F3EFDBD9' => <<-eos
+    node.default['fb_apt']['keys']['94558F59'] = nil
+    node.default['fb_apt']['keys']['F3EFDBD9'] = <<-eos
     -----BEGIN PGP PUBLIC KEY BLOCK-----
     ...
     eos
-    }
 
 Automatic key fetching can be disabled by setting the keyserver to `nil`; this 
 will produce an exception for any unspecified key. By default `fb_apt` will 
@@ -73,18 +71,22 @@ anything in `/etc/apt/apt.conf.d`. Example:
       'Acquire::http' => {
         'Proxy' => 'http://myproxy:3412',
       },
-    }
-  
+    })
+
+Internally, `fb_apt` uses the `FB::Apt.gen_apt_conf_entry()` library function
+to generate this config file. This is an internal function that's not meant to 
+be used by other cookbooks.
+
 ### Preferences
 You can fine tune which versions of packages will be selected for installation
 by tweaking APT preferences via `node['fb_apt']['preferences']`. Note that we
 clobber the contents of `/etc/apt/preferences.d` to ensure this always takes
 precedence. Example:
 
-    node.default['fb_apt']['preferences'].merge!{
+    node.default['fb_apt']['preferences'].merge!({
       'Pin dpatch package from experimental' => {
         'Package' => 'dpatch',
         'Pin' => 'release o=Debian,a=experimental',
         'Pin-Priority' => 450,
       }
-    }
+    })
--- a/fb_apt/README.md
+++ b/fb_apt/README.md
@@ -1,0 +1,88 @@
+fb_apt Cookbook
+====================
+The `fb_apt` cookbook installs and configures APT, the Debian package 
+management tool.
+
+Requirements
+------------
+
+Attributes
+----------
+* node['fb_apt']['config']
+* node['fb_apt']['keys']
+* node['fb_apt']['keyring']
+* node['fb_apt']['keyserver']
+* node['fb_apt']['mirror']
+* node['fb_apt']['preserve_sources_list_d']
+* node['fb_apt']['preferences']
+* node['fb_apt']['repos']
+* node['fb_apt']['update_delay']
+* node['fb_apt']['want_backports']
+* node['fb_apt']['want_non_free']
+* node['fb_apt']['want_source']
+
+Usage
+-----
+To install and configure APT include `fb_apt`, which will populate the 
+repository sources in `/etc/apt/sources.list` and update the package cache 
+every `node['fb_apt']['update_delay']` seconds (defaults to 86400) using the 
+`execute[apt-get update]` resource (which other cookbooks can suscribe to or 
+notify against). Note that the `node['fb_apt']['update_delay']` attribute is 
+evaluated at compile time, so you'd need to set it before including `fb_apt` if
+you'd like to customize it.
+
+### Repository sources
+By default the cookbook will setup the base distribution repos based on the
+codename (as defined in `node['lsb']['codename']`). The mirror can be
+customized with `node['fb_apt']['mirror']`; if set to `nil`, base repos will
+not be included at all in `/etc/apt/sources.list`. If base repos are enabled,
+the additional `backports` and `non-free` sources can be enabled with the
+`node['fb_apt']['want_backports']` and `node['fb_apt']['want_non_free']`
+attributes, and source code repos can be enabled with
+`node['fb_apt']['want_source']`; these all default to `false`.
+
+Additiona repository sources can be added with `node['fb_apt']['repos']`. By
+default `fb_apt` will clobber existing contents in `/etc/apt/sources.list.d` to
+ensure it has full control on the repository list; this can be disabled with
+`node['fb_apt']['preserve_sources_list_d']`.
+
+### Keys
+Repository keys can be added to `node['fb_apt']['keys']` which is a hash in the
+`keyid => key` format; if `key` is `nil` the key will be automatically fetched
+from the `node['fb_apt']['keyserver']` keyserver (`keys.gnupg.net` by default).
+Example:
+
+  node.default['fb_apt']['keys'] = {
+    '94558F59' => nil,
+    'F3EFDBD9' => <<-eos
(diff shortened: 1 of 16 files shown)


jaymzh:
Hold tight on this, it's a bunch of code and I want to review it when I have time to look. I'm catching up on life after a week of being buried in getting this all released...

fb_apt/README.md
```diff
@@ -0,0 +1,88 @@
...
+* node['fb_apt']['want_non_free']
+* node['fb_apt']['want_source']
+
+Usage
+-----
+To install and configure APT include `fb_apt`, which will populate the 
+repository sources in `/etc/apt/sources.list` and update the package cache 
+every `node['fb_apt']['update_delay']` seconds (defaults to 86400) using the 
+`execute[apt-get update]` resource (which other cookbooks can suscribe to or 
+notify against). Note that the `node['fb_apt']['update_delay']` attribute is
```
jaymzh: s/ against//

fb_apt/recipes/default.rb
```diff
@@ -0,0 +1,73 @@
...
+    action :delete
+  end
+end
+
+fb_apt_sources_list 'populate sources list' do
+  notifies :run, 'execute[apt-get update]', :immediately
+end
+
+# TODO: this should be done at runtime
+pkgcache = '/var/cache/apt/pkgcache.bin'
```
jaymzh: there is no need for update_delay to be compile-time... this can easily be made runtime ... make two resources and use only_if/no_if

jaymzh: In fact, now that I think about this, I think this is the wrong approach.

Unlike yum where getting updates is very expesnive in apt it's much less so. I'd just apt-get update on every run... that's what most systems I know of do

fb_apt/README.md
jaymzh: see comments below on update_delay

fb_apt/README.md
```diff
@@ -0,0 +1,88 @@
...
+repository sources in `/etc/apt/sources.list` and update the package cache 
+every `node['fb_apt']['update_delay']` seconds (defaults to 86400) using the 
+`execute[apt-get update]` resource (which other cookbooks can suscribe to or 
+notify against). Note that the `node['fb_apt']['update_delay']` attribute is 
+evaluated at compile time, so you'd need to set it before including `fb_apt` if
+you'd like to customize it.
+
+### Repository sources
+By default the cookbook will setup the base distribution repos based on the
+codename (as defined in `node['lsb']['codename']`). The mirror can be
```
jaymzh: does this work with sid? they sometimes internchange unstable/sid in ways that don't work everywhere... I think it'll work, just asking.

fb_apt/README.md
```diff
@@ -0,0 +1,88 @@
...
+customized with `node['fb_apt']['mirror']`; if set to `nil`, base repos will
```
jaymzh: this seems an oddly dangerous default... this is where they are by default so you'll just blow the base ones away.

fb_apt/README.md
```diff
@@ -0,0 +1,88 @@
...
+ensure it has full control on the repository list; this can be disabled with
+`node['fb_apt']['preserve_sources_list_d']`.
+
+### Keys
+Repository keys can be added to `node['fb_apt']['keys']` which is a hash in the
+`keyid => key` format; if `key` is `nil` the key will be automatically fetched
+from the `node['fb_apt']['keyserver']` keyserver (`keys.gnupg.net` by default).
+Example:
+
+  node.default['fb_apt']['keys'] = {
```
jaymzh: 4 spaces for a code block... or use triplebackticks

fb_apt/README.md
```diff
@@ -0,0 +1,88 @@
...
+Automatic key fetching can be disabled by setting the keyserver to `nil`; this 
+will produce an exception for any unspecified key. By default `fb_apt` will 
+manage the keyring at `/etc/apt/trusted.gpg`; this can be customized with
+`node['fb_apt']['keyring']`.
+
+### Configuration
+APT behaviour can be customized using `node['fb_apt']['config']`, which will be
+used to populate `/etc/apt/apt.conf`. Example:
+
+  node.default['fb_apt']['config'] = {
```
jaymzh: encouraging people to wholesale overwrite the config is bad. The example should add a config.

fb_apt/README.md
```diff
@@ -0,0 +1,88 @@
...
+    'Acquire::http' => {
+      'Proxy' => 'http://myproxy:3412',
+    },
+  }
+  
+### Preferences
+You can fine tune which versions of packages will be selected for installation
+by tweaking APT preferences via `node['fb_apt']['preferences']`. Example:
+
+  node.default['fb_apt']['preferences'] = {
```
jaymzh: even more so here were they are named

fb_apt/attributes/default.rb
```diff
@@ -0,0 +1,22 @@
+# vim: syntax=ruby:expandtab:shiftwidth=2:softtabstop=2:tabstop=2
+
+if node.debian?
+  mirror = 'http://http.debian.net/debian'
+elsif node.ubuntu?
+  mirror = 'http://archive.ubuntu.com/ubuntu'
+end
```
jaymzh: wanna throw mint in here while you're at it? It's forked form ubuntu and uses apt and should just work..

fb_apt/attributes/default.rb
```diff
@@ -0,0 +1,22 @@
+# vim: syntax=ruby:expandtab:shiftwidth=2:softtabstop=2:tabstop=2
+
+if node.debian?
+  mirror = 'http://http.debian.net/debian'
```
jaymzh: that's a 404. did you mean http.us.debian.org ?

fb_apt/providers/keys.rb
```diff
@@ -0,0 +1,57 @@
...
+
+use_inline_resources
+
+action :run do
+  keyring = node['fb_apt']['keyring']
+  keyserver = node['fb_apt']['keyserver']
+  keys = node['fb_apt']['keys'].to_hash
+
+  if keys && keyring
+    cmd = Mixlib::ShellOut.new("LANG=C apt-key --keyring #{keyring} list")
```
jaymzh: lang?

fb_apt/providers/keys.rb
```diff
@@ -0,0 +1,57 @@
...
+    cmd.run_command
+    # Note: we deliberately ignore errors here, as this will fail if the keyring
```
jaymzh: that's silly, why not gaurd it with `if File.exists?(keyring)`

fb_apt/providers/sources_list.rb
```diff
@@ -0,0 +1,70 @@
...
+action :run do
+  repos = []
+  mirror = node['fb_apt']['mirror']
+  distro = node['lsb']['codename']
+
+  # only add base repos if mirror is set and codename is available
+  if mirror && distro
+    components = 'main'
+    if node.ubuntu?
+      components += ' universe'
```
jaymzh: isn't universe like contrib ? does that need to be configurable?

fb_apt/providers/sources_list.rb
```diff
@@ -0,0 +1,70 @@
...
+use_inline_resources
+
+action :run do
+  repos = []
+  mirror = node['fb_apt']['mirror']
+  distro = node['lsb']['codename']
+
+  # only add base repos if mirror is set and codename is available
+  if mirror && distro
+    components = 'main'
```
jaymzh: it's an list, so instead of managing spaces manually which will break at some point with some diff, make it an array, and when you use it, call join

fb_apt/providers/sources_list.rb
```diff
@@ -0,0 +1,70 @@
...
+    # Security updates
+    if node.debian?
+      base_repos <<
+        "http://security.debian.org/ #{distro}/updates #{components}"
+    elsif node.ubuntu?
+      base_repos <<
+        "http://security.ubuntu.com/ #{distro}-security #{components}"
+    end
+
+    # Stable updates
```
jaymzh: these don't exist for sid...

fb_apt/recipes/default.rb
```diff
@@ -0,0 +1,73 @@
...
+  source 'apt.conf.erb'
+  owner 'root'
+  group 'root'
+  mode '0644'
+  notifies :run, 'execute[apt-get update]'
+end
+
+# No sane package should drop stuff here, and bad preferences can seriously
+# mess up a machine, so let's clobber it to be safe.
+Dir.glob('/etc/apt/preferences.d/*').each do |f|
```
jaymzh: you lack an option here, unlike in sources.list.d and should probably document that

fb_apt/templates/default/apt.conf.erb
```diff
@@ -0,0 +1,30 @@
+// This file is maintained by Chef. Do not edit, all changes will be
+// overwritten. See fb_apt/README.md
+<%
+  def make_option(k, v, i)
```
jaymzh: put this in a library file and scope it to FB::Apt

jaymzh:
In addition you need the various boilerplate in every file, please.

fb_apt/recipes/default.rb
davide125: I wouldn't want to do it on every run for resource constrained systems, or if the cache is on flash, as it writes out a 10s of MB every time.

fb_apt/README.md
davide125: It should, but I'll test that, good point.

fb_apt/README.md
davide125: Well the idea is to takeover sources.list with a known good set. We do give the option to opt out though.

fb_apt/README.md
davide125: yup, good point

fb_apt/attributes/default.rb
davide125: sure I'll need the default mirror url and the node['platform'] entry for it

fb_apt/attributes/default.rb
davide125: nope, and the 404 is expected. http://http.debian.net is a magic redirecting mirror that sends you to the closest one to you.

fb_apt/providers/keys.rb
davide125: apt is localized, this is to ensure we can parse the output reliably

fb_apt/providers/keys.rb
davide125: because I was lazy and this is annoying to test :) I'll rework this

fb_apt/providers/sources_list.rb
davide125: So my understanding is that universe is like main, but the packages are not necessarily officially supported by ubuntu. Every single ubuntu system I've seen in the wild has it enabled, and the license is the same as main, so IMO it makes sense to enable it here.

fb_apt/providers/sources_list.rb
davide125: oh I see what you mean. Yeah this makes sense, I'll change it to use an array.

fb_apt/providers/sources_list.rb
davide125: good catch, I'll gate them out for it

fb_apt/recipes/default.rb
davide125: yeah I went back and forth on this, but I don't think there's any legit reason to leave stuff there unmanaged. I'll add a note to the README about it

fb_apt/templates/default/apt.conf.erb
davide125: will do

fb_apt/attributes/default.rb
davide125: Looked into this, it's actually non-trivial as mint is layered on top of ubuntu (they don't provide full mirrors for it), so we'd need some logic to map the mint -> ubuntu codenames. I'll leave this out for now, but feel free to add it in a separate PR if you want.

fb_apt/recipes/default.rb
davide125: README updated

fb_apt/recipes/default.rb
jaymzh: it shouldn't. modern debian uses delta metadata and it only downloads new pdiff files... it's designed to be run regularly.

fb_apt/README.md
jaymzh: no I think you misunderstand. I'm saying that if the person sets the mirror to 'nil'...  mkaing their system unusable seems bad... especially because by defualt we also blow-away sources.d

But given your default isn't nil, it makes sense. UPdate to say that the default is a global mirror for wahtever your distro is

fb_apt/attributes/default.rb
jaymzh: that page says specifically you should use httpredir.debian.org/debian - which it turns out is NOT a 404. :)

fb_apt/providers/keys.rb
jaymzh: why not en_US.UTF-8 ? C seems crazy

fb_apt/README.md
```diff
@@ -0,0 +1,92 @@
...
+* node['fb_apt']['update_delay']
+* node['fb_apt']['want_backports']
+* node['fb_apt']['want_non_free']
+* node['fb_apt']['want_source']
+
+Usage
+-----
+To install and configure APT include `fb_apt`, which will populate the 
+repository sources in `/etc/apt/sources.list` and update the package cache 
+every `node['fb_apt']['update_delay']` seconds (defaults to 86400) using the
```
jaymzh: see other comments. I think every run is fine. Also, if you do leave this feature the description is misleading. It's not every `update_delay` seconds, it's every run where the MD is older than `update_delay` seconds

fb_apt/README.md
```diff
@@ -0,0 +1,92 @@
...
+used to populate `/etc/apt/apt.conf`. Note that this will take precedence over
+anything in `/etc/apt/apt.conf.d`. Example:
+
+    node.default['fb_apt']['config'].merge!({
+      'Acquire::http' => {
+        'Proxy' => 'http://myproxy:3412',
+      },
+    })
+
+Internally, `fb_apt` uses the `FB::Apt.gen_apt_conf_entry()` library function
```
jaymzh: call it `_gen_apt_conf_entry` if that's the case

fb_apt/attributes/default.rb
jaymzh: fair enough

fb_apt/providers/keys.rb
```diff
@@ -0,0 +1,69 @@
...
+    if ::File.exists?(keyring)
+      cmd = Mixlib::ShellOut.new("LANG=C apt-key --keyring #{keyring} list")
+      cmd.run_command
+      cmd.error!
+      output = cmd.stdout.split("\n")
+      Chef::Log.debug("apt-key output: #{output.join("\n")}")
+      installed_keys = output.select { |x| x.start_with?('pub') }.map do |x|
+        x[%r/pub.*\/(?<keyid>[A-Z0-9]*)/, 'keyid']
+      end
+      Chef::Log.info("Installed keys: #{installed_keys.join(', ')}")
```
jaymzh: I'm fine with this code, but it came be made a bit simpler with:
```
installed_keys = []
if File.exists?(keyring)
  ...
  installed_keys = ...
end
Chef::Log.debug("Installed keys: ...")
```

No need for different print statements, you have a default, etc.

Also, that should be debug.

fb_apt/providers/sources_list.rb
```diff
@@ -0,0 +1,79 @@
...
+        components += %w{restricted multiverse}
+      else
+        fail "Don't know how to setup non-free for #{node['platform']}"
+      end
+    end
+
+    components_entry = components.join(' ')
+    base_repos = [
+      # Main repo
+      "#{mirror} #{distro} #{components_entry}",
```
jaymzh: `components.join(' ')` works in `#{}` and fits on that line just fine, fwiw

fb_apt/providers/sources_list.rb
```diff
@@ -0,0 +1,79 @@
...
+    # Security updates
+    if node.debian? && distro != 'sid'
+      base_repos <<
+        "http://security.debian.org/ #{distro}/updates #{components_entry}"
+    elsif node.ubuntu?
+      base_repos <<
+        "http://security.ubuntu.com/ #{distro}-security #{components_entry}"
+    end
+
+    # Stable updates
```
jaymzh: again, sid doesn't have updates.

fb_apt/providers/sources_list.rb
```diff
@@ -0,0 +1,79 @@
...
+
+  # add custom repos
+  repos += node['fb_apt']['repos']
+
+  template '/etc/apt/sources.list' do
+    source 'sources.list.erb'
+    owner 'root'
+    group 'root'
+    mode '0644'
+    variables(:repos => repos)
```
jaymzh: don't use variables, please

fb_apt/recipes/default.rb
davide125: The pdiff stuff is only for download, it will still rewrite the entire pkgcache every time, even if just one of the deltas changed. You do not want to do this on a small box with limited flash endurance.

fb_apt/README.md
davide125: Yeah I updated the README in the latest revision to make this clearer. The idea behind `nil` is that in some cases you might prefer having full control on the sources and not rely on the structure of the default mirrors. In that case you'd set this to `nil` and pass every repo you need in `node['fb_apt']['repos']`.

fb_apt/attributes/default.rb
davide125: ah, good catch, will fix this

fb_apt/providers/keys.rb
davide125: Because you won't necessarily have en_US.UTF-8 on a box (you can disable it if you dpkg-reconfigure the locales package or mess with localegen). C is guaranteed to be everywhere (it's a POSIX thing actually) and give you plain output in English, so it's a good choice for parsing.

fb_apt/providers/sources_list.rb
davide125: I know, didn't want to repeat the join on every entry, this looks cleaner IMO.

fb_apt/providers/sources_list.rb
davide125: what should I use instead? I'd rather keep the mirror generation logic in the provider than move it to the template, that's why I'm passing a variable here (as repos is mirrors stuff + node['fb_apt']['repos'])

fb_apt/providers/sources_list.rb
jaymzh: didn't notice the repeat. good call.

fb_apt/providers/sources_list.rb
jaymzh: node.default['fb_apt']['repos] = repos + node['fb_apt']['repos']

then access the node directly from the template.

fb_apt/README.md
```diff
@@ -25,9 +25,9 @@ Usage
 -----
 To install and configure APT include `fb_apt`, which will populate the 
 repository sources in `/etc/apt/sources.list` and update the package cache 
-every `node['fb_apt']['update_delay']` seconds (defaults to 86400) using the 
-`execute[apt-get update]` resource (which other cookbooks can suscribe to or 
-notify).
+during the run if it's old than `node['fb_apt']['update_delay']` seconds
```
jaymzh: s/old/older

fb_apt/README.md
```diff
@@ -25,9 +25,9 @@ Usage
...
+(defaults to 86400). This uses the `execute[apt-get update]` resource, which
+other cookbooks can suscribe to or notify as well.
```
jaymzh: it's obvious if you read the code, but you should note that setting this to 0 is the "right" way to have it update on every run

jaymzh:
OK the code looks good to me. Have you tested all the changes on both Debian and Ubuntu?

davide125:
I've tested this extensively on Debian. I haven't tested on Ubuntu yet, but given that the tools and mechanics are the same it should just work there.

jaymzh:
Can you try it in a VM or something? Just like the hostname was wrong debian I'd like to make sure it at least *passingly* works on Ubuntu since we claim to support it.

davide125:
Ok I got this running in an Ubuntu 14.04 VM as well, apart from a minor issue with signing keys (fixed in the last commit) it works fine.

jaymzh:
Aren't ya glad ya tested? :)

OK then we're all good. New cookbooks require some finagling to get ShipIt to do the right thing so I'll merge this when I'm back in the office on Monday.

Thanks!

jaymzh:
@facebook-github-bot shipit

PR #3, opened 2016-05-01
add fb_dnsmasq cookbook

davide125:
This adds a cookbook to manage dnsmasq using the attribute-driven API. It's a rework of an unpublished cookbook I've been running at home for a while. Tested only on Debian, but it should work fine on other distros as there's nothing Debian-specific in it.

### DIFF
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -34,4 +34,6 @@ provisioner:
 
 suites:
   - name: default
-    run_list: recipe[fb_init_sample]
+    run_list:
+      - recipe[fb_init_sample]
+      - recipe[fb_dnsmasq]
--- a/cookbooks/fb_dnsmasq/README.md
+++ b/cookbooks/fb_dnsmasq/README.md
@@ -1,0 +1,40 @@
+fb_dnsmasq Cookbook
+====================
+This cookbook installs and configures dnsmasq, a small caching DNS proxy and
+DHCP/TFTP server.
+
+Requirements
+------------
+
+Attributes
+----------
+* node['fb_dnsmasq']['config']
+
+Usage
+-----
+Include `fb_dnsmasq` in your runlist to install dnsmasq and enable its service.
+Configuration can be customized using `node['fb_dnsmasq']['config']` according
+to the [upstream documentation](http://www.thekelleys.org.uk/dnsmasq/docs/dnsmasq-man.html). Example:
+
+    node.default['fb_dnsmasq']['config'] = {
+      'dhcp-authoritative' => nil,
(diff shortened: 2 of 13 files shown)


fb_dnsmasq/recipes/default.rb
```diff
@@ -0,0 +1,41 @@
...
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree. An additional grant
+# of patent rights can be found in the PATENTS file in the same directory.
+#
+
+package 'dnsmasq' do
+  action :upgrade
+end
+
+template '/etc/ethers' do
```
jaymzh: oh this seems... *terrible*. /etc/ethers is a real thing, and while having dnsmasq read it is one thing, this cookbook CERTAINLY shouldn't write it. Feel free to make an fb_ethers akin to fb_hosts...  but certainly just because I don't want dnsmasq to USE /etc/ethers doesn't mean I want to DELETE it. other things use it

davide125: Ah interesting, I did not know this was a general purpose thing, but you're right. I'll spin this out into a `fb_ethers` cookbook.

.kitchen.yml
```diff
@@ -34,4 +34,6 @@ provisioner:
 
 suites:
   - name: default
-    run_list: recipe[fb_init_sample]
+    run_list:
```
jaymzh: you're missing a "  - " here

fb_dnsmasq/attributes/default.rb
```diff
@@ -0,0 +1,13 @@
...
+#
+# Copyright (c) 2016-present, Facebook, Inc.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree. An additional grant
+# of patent rights can be found in the PATENTS file in the same directory.
+#
+
+default['fb_dnsmasq'] = {
```
jaymzh: at the very least you need an enable/disable option here.

jaymzh:
you also need a rebase.

.kitchen.yml
davide125: nope, kitchen is weird: see https://docs.chef.io/config_yml_kitchen.html

goncalopereira: oh yes name and run_list are both at the same level inside the list (list item start represented by the '-') https://en.wikipedia.org/wiki/YAML#Basic_components_of_YAML

In Json it would be something like:

`[{name: 'default',run_list: ['recipe[fb_init_sample]']}]`

PR #4, opened 2016-05-01
add fb_p910nd cookbook

davide125:
This adds a cookbook to manage p910nd using the attribute-driven API. It's a rework of an unpublished cookbook I've been running at home for a while. Tested only on Debian, but it should work fine on Ubuntu too as the package there is exactly the same.

### DIFF
--- a/fb_p910nd/README.md
+++ b/fb_p910nd/README.md
@@ -1,0 +1,24 @@
+fb_p910d Cookbook
+====================
+This cookbook installs and configures p910nd, a lightweight print server.
+
+Requirements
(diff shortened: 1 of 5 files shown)


davide125:
Not sure why foodcritic is failing here:
`FC019: Access node attributes in a consistent manner: fb_p910nd/attributes/default.rb:11`

jaymzh:
This doesn't fall into the scope of what we keep in this repo. Instead, feel free to send a pullrequest to UNIVERSE.md to point to your own repo.

PR #23, opened 2016-09-16
Undo recent fixes that are breaking syncing.

jaymzh:
Will redo this for the sync.


jaymzh:
This will break tests... I know.

PR #30, opened 2016-10-26
Adding few more node functions

rameshshihora:
Please review.


jaymzh:
@rameshshihora  - your tests are failing...

but you should make this diff internally anyway, that way we don't have to import/export it for you.

PR #32, opened 2016-11-09
Remove reference to oncalls from public doc

elthariel:
Hi oscore :),

Was just browsing the repo, considering using some of this stuff for my own infra and came accross a reference to the chef oncall, which probably shouldn't be laying around open source code.

Signed-off-by: Julien 'Lta' BALLET <contact@lta.io>

### DIFF
--- a/cookbooks/fb_fstab/README.md
+++ b/cookbooks/fb_fstab/README.md
@@ -35,7 +35,6 @@ add whatever filesystems you would like mounted and `fb_fstab` will, for each
 entry in the hash:
 * populate `/etc/fstab` for you
 * create the `mount_point` (but not parents) if it doesn't exist
-  (if you need parents created, please file a task against chef oncall)
 * mount the filesystem
 
 ### Global Options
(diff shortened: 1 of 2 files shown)


jaymzh:
It's telling you that because you're an ex-employee and thus the tool can't overwrite the existing CLA with a new one. Which requires sending in a physically signed paper - cla@ can help, but since this is simply 2 line comment change, we'll call it "obvious fix" and merge it.

elthariel:
Yeah, I assumed it was something along those lines, I've send them an email.

Thx

jaymzh:
OK I was an idiot and merged this here instead of telling the bot to do it and it broke things so I had to revert it. I'm going to inquire about if I can make the bot pull in a closed PR, but if you wanna send another PR I'll do it right... and/or I can just make a PR and say @elthariel  was the original author...

jaymzh:
Alright I was able to make a diff that got sync'd as you. We're all good. :)

elthariel:
Haha, shit happens :) but don't worry about it, it's just 2 lines.

PR #33, opened 2016-12-08
Determine configuration and sites folder paths based on Apache versio…

skywalk7:
Major changes:
1. Support for Apache 2.4 on Debian platform, so we can put configuration and sites into correct folder.
2. Default sites sometimes conflicts with applying configuration, so adding flag to delete those.

### DIFF
--- a/cookbooks/fb_apache/attributes/default.rb
+++ b/cookbooks/fb_apache/attributes/default.rb
@@ -105,6 +105,7 @@ default['fb_apache'] = {
     'rewrite' => 'mod_rewrite.so',
     'setenvif' => 'mod_setenvif.so',
     'speling' => 'mod_speling.so',
+    'ssl' => 'mod_ssl.so',
     'status' => 'mod_status.so',
     'substitute' => 'mod_substitute.so',
     'suexec' => 'mod_suexec.so',
--- a/cookbooks/fb_apache/recipes/default.rb
+++ b/cookbooks/fb_apache/recipes/default.rb
@@ -10,20 +10,51 @@
 # of patent rights can be found in the PATENTS file in the same directory.
 #
 
-confdir = value_for_platform_family(
-  'rhel' => '/etc/httpd/conf.d',
-  'debian' => '/etc/apache2/conf.d',
-)
+apache_version =
+  case node['platform_family']
+  when 'debian'
+    case node['platform']
+    when 'ubuntu'
+      node['platform_version'].to_f >= 13.10 ? '2.4' : '2.2'
+    when 'debian'
+      node['platform_version'].to_f >= 8.0 ? '2.4' : '2.2'
+    else
+      '2.4'
+    end
+  when 'rhel'
+    node['platform_version'].to_f >= 7.0 ? '2.4' : '2.2'
+  end
+
+confdir =
+  case node['platform_family']
+  when 'rhel'
+    '/etc/httpd/conf.d'
+  when 'debian'
+    case apache_version
+    when '2.2'
+      '/etc/apache2/conf.d'
+    when '2.4'
+      '/etc/apache2/conf-enabled'
+    end
+  end
 
 sitesdir = value_for_platform_family(
   'rhel' => confdir,
   'debian' => '/etc/apache2/sites-enabled',
 )
 
-moddir = value_for_platform_family(
-  'rhel' => '/etc/httpd/conf.modules.d',
-  'debian' => '/etc/apache2/modules-enabled',
-)
+moddir =
+  case node['platform_family']
+  when 'rhel'
+    '/etc/httpd/conf.modules.d'
+  when 'debian'
+    case apache_version
+    when '2.2'
+      '/etc/apache2/modules-enabled'
+    when '2.4'
+      '/etc/apache2/mods-enabled'
+    end
+  end
 
 sysconfig = value_for_platform_family(
   'rhel' => '/etc/sysconfig/httpd',


cookbooks/fb_apache/recipes/default.rb
```diff
@@ -81,6 +97,16 @@
   notifies :restart, 'service[apache]'
 end
 
+# Remove default sites
+node['fb_apache']['default_sites'].each do |site|
+  file "#{sitesdir}/#{site}" do
+    action :delete
+    force_unlink true
+    notifies :restart, 'service[apache]'
+    not_if node['fb_apache']['default_sites_enabled']
```
jaymzh: That'll get interpreted as a string and thus always true. `not_if { ...}`

cookbooks/fb_apache/recipes/default.rb
```diff
@@ -81,6 +97,16 @@
   notifies :restart, 'service[apache]'
 end
 
+# Remove default sites
+node['fb_apache']['default_sites'].each do |site|
```
jaymzh: This isn't runtime safe and doesn't follow our model. If people can't modify it after this cookbook, it won't fly.

cookbooks/fb_apache/recipes/default.rb
```diff
@@ -10,20 +10,36 @@
...
-confdir = value_for_platform_family(
-  'rhel' => '/etc/httpd/conf.d',
-  'debian' => '/etc/apache2/conf.d',
-)
+confdir =
+  case node['platform_family']
+  when 'rhel'
+    '/etc/httpd/conf.d'
+  when 'debian'
+    case node['fb_apache']['version']
```
jaymzh: This isn't a runtime-safe API.

cookbooks/fb_apache/attributes/default.rb
```diff
@@ -18,6 +33,15 @@
 auth_core_suffix = node.centos6? ? 'default' : 'core'
 
 default['fb_apache'] = {
+  'default_sites_enabled' => true,
+  'default_sites' => [
```
jaymzh: I don't think we should do this at all. The whole point of the model we have is all the sites are in one file managed by the cookbook. We should wholesale remove everything from sites-enabled other than our file.

cookbooks/fb_apache/attributes/default.rb
```diff
@@ -9,6 +9,21 @@
 # of patent rights can be found in the PATENTS file in the same directory.
 #
 
+default['fb_apache']['version'] =
```
jaymzh: this wont' work - we overwrite the entire hash below.

cookbooks/fb_apache/attributes/default.rb
skywalk7: I agree. What is the best way to purge unmanaged files from directory? remote_directory doesn't work because of template...

jaymzh: I'd do:

```
Dir.new(sites_dir).each do |entry|
  next if ['.', '..', 'fb-sites.cfg'].include?(entry)
  file entry do
    action :delete
  end
end
```

jaymzh: (psuedo-code, haven't tested it)

cookbooks/fb_apache/metadata.rb
```diff
@@ -4,6 +4,7 @@
 license 'BSD'
 description 'Installs/Configures Apache'
 source_url 'https://github.com/facebook/chef-cookbooks/'
+issues_url 'https://github.com/facebook/chef-cookbooks/issues'
```
jaymzh: please keep this consistent with the rest of our cookbooks. Since issues URL is consistent with the source_url we have only defined one.

cookbooks/fb_apache/recipes/default.rb
```diff
@@ -81,6 +112,16 @@
   notifies :restart, 'service[apache]'
 end
 
+# Remove unmanaged sites
+::Dir.glob("#{sitesdir}/*").reject {|e| ::File.directory?(e)}.each do |entry|
+  next if ['fb_sites.conf', 'fb_apache.conf'].include?(::File.basename(entry))
```
jaymzh: We can't do this if confdir and sitesdir are the same because we'll nuke the ssl.conf and php.conf and stuff that comes with the package and sets some sane defaults... doing this on Debian is clearly a good call though.

cookbooks/fb_apache/metadata.rb
skywalk7: I've missed `scripts/run_foodcritic` and fixed foodcritic warning. I will rollback change.

cookbooks/fb_apache/recipes/default.rb
jaymzh: You don't, but you do need:

```
Listen 443 https
SSLSessionCache         shmcb:/run/httpd/sslcache(512000)
SSLSessionCacheTimeout  300
SSLRandomSeed startup file:/dev/urandom  256
SSLRandomSeed connect builtin
SSLCryptoDevice builtin
```

Or for php:

```
<FilesMatch \.php$>
    SetHandler application/x-httpd-php
</FilesMatch>
AddType text/html .php
DirectoryIndex index.php
php_value session.save_handler "files"
php_value session.save_path    "/var/lib/php/session"
```

Without which PHP basically doesn't work... and the distro keeps this stuff up-to-date. @davide125  and I had talked about nuking this stuff and making people put it in their `node['fb_apache']['extra_configs']` but it seemed a bit excessive...

skywalk7:
I've removed controversial part, so we can move forward. There should be better non-disruptive and uniform approach to the issue.

jaymzh:
Will import/test/land. It'll get sync'd back out as you.

PR #36, opened 2017-11-07
Added archlinux support

zfjagann:
1. Added a systemd package list for arch, so the packages can be managed by this cookbook, instead of failing.
1. Added attributes `node['fb_systemd']['networkd']['ignore']` and `node['fb_systemd']['resolved']['ignore']` to prevent this cookbook from taking any action on the networkd/resolved resources. [Example of this in action](https://gist.github.com/zfjagann/507e3b48a78350e716119e962b496292)

### DIFF
--- a/cookbooks/fb_systemd/README.md
+++ b/cookbooks/fb_systemd/README.md
@@ -22,8 +22,10 @@ Attributes
 * node['fb_systemd']['logind']['enable']
 * node['fb_systemd']['logind']['config']
 * node['fb_systemd']['networkd']['enable']
+* node['fb_systemd']['networkd']['ignore']
 * node['fb_systemd']['resolved']['enable']
 * node['fb_systemd']['resolved']['config']
+* node['fb_systemd']['resolved']['ignore']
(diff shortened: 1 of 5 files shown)


jaymzh:
Unfortuantely, this violates the approach and model of Facebook's core cookbooks. The arch support is welcome, but we explicitly don't ignore stuff - our cookbooks own an entire system or leave it alone. What other configuration knobs would you need to just manage that from Chef?

zfjagann:
Then provide a knob to disable this recipe (which I think is difficult to do). It's annoying to have the cookbook either enable the recipes, or disable them, with no option to just leave them as-is.

davide125:
Why do you want to leave resolved/networkd unmanaged? As @jaymzh said the general philosophy is for a cookbook to manage everything; if the current API for resolved/networkd doesn't work for your usecase we should fix that, rather than adding knobs to skip parts of the recipe.

davide125:
I'm going to close this as it's been inactive for almost two years.

PR #50, opened 2019-01-30
Ensure attributes for fb_timers only runs on supported OSes

natewalck:
Currently, if fb_timers is in a `depends` on a cookbook but *not* included into the runlist, it will fail on any OS where it isn't supported.

This diff ensures that the attributes only run on OSes where it is intended to run.

### DIFF
--- a/cookbooks/fb_timers/attributes/default.rb
+++ b/cookbooks/fb_timers/attributes/default.rb
@@ -8,6 +8,8 @@
 # of patent rights can be found in the PATENTS file in the same directory.
 #
 
+return unless node.centos? || node.debian? || node.ubuntu?
+
 default['fb_timers'] = {
   'jobs' => {},
 


natewalck:
You are correct on those points.

I just realized the error I saw was due to a previous version pulled a month or two ago, which had this line:

```
  '_fix_timer_wantedby' => node.in_shard?(39),
```

This relied upon `node.in_shard?` from fb_helpers. 

In fb_helpers, it is assumed that either node['shard_seed'] or node['fb']['shard'] exists. This was blowing up on our machines because we had neither. 

I'll sync up versions, but perhaps in the future it is worth looking for sharded deployments and making sure they don't hit the opensource version or not make assumptions about which node attributes are present without providing them.

jaymzh:
`node['shard']` is provided from upstream Ohai... you should have it...

natewalck:
@jaymzh As of which version of chef?
```
chef (14.8.12)> node['shard']
 => nil
chef (14.8.12)> node['shard_seed']
 => nil
```

natewalck:
Also ohai directly:
```
sh-3.2# ohai shard_seed
Traceback (most recent call last):
	9: from /usr/local/bin/ohai:306:in `<main>'
	8: from /usr/local/bin/ohai:306:in `load'
	7: from /opt/chefdk/embedded/lib/ruby/gems/2.5.0/gems/ohai-14.8.10/bin/ohai:26:in `<top (required)>'
	6: from /opt/chefdk/embedded/lib/ruby/gems/2.5.0/gems/ohai-14.8.10/lib/ohai/application.rb:80:in `run'
	5: from /opt/chefdk/embedded/lib/ruby/2.5.0/benchmark.rb:293:in `measure'
	4: from /opt/chefdk/embedded/lib/ruby/gems/2.5.0/gems/ohai-14.8.10/lib/ohai/application.rb:82:in `block in run'
	3: from /opt/chefdk/embedded/lib/ruby/gems/2.5.0/gems/ohai-14.8.10/lib/ohai/application.rb:114:in `run_application'
	2: from /opt/chefdk/embedded/lib/ruby/gems/2.5.0/gems/ohai-14.8.10/lib/ohai/application.rb:114:in `each'
	1: from /opt/chefdk/embedded/lib/ruby/gems/2.5.0/gems/ohai-14.8.10/lib/ohai/application.rb:115:in `block in run_application'
/opt/chefdk/embedded/lib/ruby/gems/2.5.0/gems/ohai-14.8.10/lib/ohai/system.rb:178:in `attributes_print': I cannot find an attribute named shard_seed! (ArgumentError)
```

jaymzh:
Sometime in 12. But in 14 they turned a bunch of plugins into "optional" plugins, and shard was one of the casualities... which means you have to turn it back on... it was in the release notes for 14. :)

Issue #52, opened 2019-02-21
apt-key does not output short key IDs in debian 9


chrisrlong:
Hi,

The provider code:

 if keys && keyring
    installed_keys = []
    if ::File.exist?(keyring)
      cmd = Mixlib::ShellOut.new("LANG=C apt-key --keyring #{keyring} list")
      cmd.run_command
      cmd.error!
      output = cmd.stdout.split("\n")
      Chef::Log.debug("apt-key output: #{output.join("\n")}")
      installed_keys = output.select { |x| x.start_with?('pub') }.map do |x|
        **x[%r{pub.*/(?<keyid>[A-Z0-9]*)}**, 'keyid']
      end
    end

Does not detect the debian 9 apt-key output:

pub   rsa4096 2016-10-05 [SC]
      72EC F46A 56B4 AD39 C907  BBB7 1646 B01B **86E5 0310**
uid           [ unknown] Yarn Packaging <yarn@dan.cx>

Every chef pass reinstalls all the keys.

chrisrlong:
please see PR #53

malmond77:
Thanks for the report, and the PR. Let's follow up there.

PR #53, opened 2019-02-22
Updating keys.rb to use long format fingerprint

chrisrlong:
Updating the regex in keys.rb to use match on long format fingerprint to get the key ID as the short form is removed in newer versions of apt-key.


chrisrlong:
this is to address https://github.com/facebook/chef-cookbooks/issues/52

chrisrlong:
Travis failure due to an ssh timeout, not this code.

davide125:
Restarted failed the Travis job

PR #59, opened 2019-08-16
New cookbook: fb_users

jaymzh:
This cookbook is used to manage users via the FB attribute-driven model.

Adding users or groups is simply done in the node as you'd expect. The
README has copious documentation.

This has one significant departure from how most FB-style cookbooks
work, and that is that is it has users monkeypatch a class. The reason
for this is as follows. One important piece of a cookbook that managers
users is to enforce consistent UIDs and GIDs across the environment,
which means it should not be part of the API. As I recall, at FB these
exist in an attribute that "shouldn't be touched", and I wanted to keep
that general idea but formalize it better. Putting it in attributes
would mean users *always* pass in UIDs via the node which just begs for
bad behavior. So I decided to put it in a class constant.

The nice thing about this is that as long as the cookbook with the
monkeypatched library is *depended on* by anything in the base runlist,
from fb_users to fb_init to whatever, it all just works.

One other small change from how I recall FB doing this internally... the
UID/GID map can also include the user/group "comment" since that likely
also should be consistent across the world.

### DIFF
--- a/cookbooks/fb_users/resources/default.rb
+++ b/cookbooks/fb_users/resources/default.rb
@@ -30,24 +30,33 @@ action :manage do
     "#{pgroups.join(', ')}.",
   )
   pgroups.each do |grp|
-    if node['etc']['group'][grp]
+    if node['etc']['group'][grp] &&
+        node['etc']['group'][grp]['gid'] == ::FB::Users::GID_MAP[grp]['gid']
       Chef::Log.debug(
-        "fb_users: Will not bootstrap group #{grp} since it exists",
+        "fb_users: Will not bootstrap group #{grp} since it exists, and has " +
+        'the right GID',
       )
       next
     end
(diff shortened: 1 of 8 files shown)


jaymzh:
Updated with a small group bugfix and license headers.

cookbooks/fb_users/resources/default.rb
```diff
@@ -0,0 +1,134 @@
...
+
+  # and then converge all groups
+  node['fb_users']['groups'].each do |groupname, info|
+    if info['action'] == :delete
+      group groupname do
+        action :remove
+      end
+      next
+    end
+    # disableing fc009 becasue it triggers on 'comment' below which
```
davide125: s/disableing/disabling/
s/becasue/because/

jaymzh:
rebased at @NaomiReeves request

PR #60, opened 2019-08-16
New cookbook: fb_sudo

jaymzh:
This cookbook manages `/etc/sudoers` using an FB-style attribute-driven
API.

### DIFF
--- a/cookbooks/fb_sudo/attributes/default.rb
+++ b/cookbooks/fb_sudo/attributes/default.rb
@@ -1,0 +1,40 @@
+#
+# Copyright (c) 2019-present, Vicarious, Inc.
+# All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+default['fb_sudo'] = {
+  'aliases' => {
+    'host' => {},
+    'user' => {},
+    'command' => {},
+    'runas' => {},
+  },
+  'defaults' => {
+    'visiblepw' => false,
+    'always_set_home' => true,
+    'env_reset' => true,
+    'env_keep' => 'COLORS DISPLAY HOSTNAME HISTSIZE INPUTRC KDEDIR LS_COLORS ' +
+      'MAIL PS1 PS2 QTDIR USERNAME LANG LC_ADDRESS LC_CTYPE ' +
+      'LC_COLLATE LC_IDENTIFICATION LC_MEASUREMENT LC_MESSAGES ' +
+      'LC_MONETARY LC_NAME LC_NUMERIC LC_PAPER LC_TELEPHONE ' +
+      'LC_TIME LC_ALL LANGUAGE LINGUAS _XKB_CHARSET XAUTHORITY',
+    'secure_path' => '/sbin:/bin:/usr/sbin:/usr/bin',
+  },
+  'default_overrides' => {},
+  'users' => {
+    '%sudo' => 'ALL=(ALL) ALL',
+  },
+}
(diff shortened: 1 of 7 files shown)


cookbooks/fb_sudo/attributes/default.rb
```diff
@@ -0,0 +1,23 @@
...
+    'always_set_home' => true,
+    'env_reset' => true,
+    'env_keep' => 'COLORS DISPLAY HOSTNAME HISTSIZE INPUTRC KDEDIR LS_COLORS ' +
+      'MAIL PS1 PS2 QTDIR USERNAME LANG LC_ADDRESS LC_CTYPE ' +
+      'LC_COLLATE LC_IDENTIFICATION LC_MEASUREMENT LC_MESSAGES ' +
+      'LC_MONETARY LC_NAME LC_NUMERIC LC_PAPER LC_TELEPHONE ' +
+      'LC_TIME LC_ALL LANGUAGE LINGUAS _XKB_CHARSET XAUTHORITY',
+    'secure_path' => '/sbin:/bin:/usr/sbin:/usr/bin',
+  },
+  'default_overrides' => {},
```
davide125: I think we should probably make the `default` / `default_overrides` API a bit more specific -- as it is, it's going to be confusing to use unless one remembers the modifiers by heart. How about
```
defaults' => {
  'generic' => { ...},
  'host' => {},
  'user' => {},
  'command' => {},
  'runas' => {},
}
```

Also, I hadn't realized till now that the ordering of these actually matters:
"
     Defaults entries are parsed in the following order: generic, host, user
     and runas Defaults first, then command defaults.  If there are multiple
     Defaults settings of the same type, the last matching setting is used.
     The following Defaults settings are parsed before all others since they
     may affect subsequent entries: fqdn, group_plugin, runas_default,
     sudoers_locale.
"

so we may wanna make these arrays and not hashes.

jaymzh: I"m not sure it makes sense to build empty hashes there, it doesn't change behavior... people can always add new entries, and because of autovivification, it doesn't matter if we've created an empty array or not. Plus  what about `fdqn`, `plugin`, etc.?  And in fact they've added some over time, so we don't want to play that game, like anything else in the fb-chef world, there's a 1:1 mapping, and the `sudoers` man page is the docs for this, so we're not re-creating it.

As to arrays vs. hashes - arrays are ordered in Ruby (as of 1.9, and we're on 2.4 or 2.5 depending on your version of Chef), and thus having names so you can modify it later in the run outweighs the value of having it be an array, IMO.

davide125: My point is mostly that with the current API one ends up having to write stuff like
```
node.default['fb_sudo']['default_overrides'][':root']['passwd_tries'] = 0
```
which isn't the most intuitive, as I wouldn't expect people to necessarily remember the modifiers. The split between `defaults` and `default_overrides` is also pretty arbitrary -- these aren't framed as "overrides" in the sudo documentation, they're just scoped defaults.

jaymzh: No, they are literal overrides, per the man page:

```
     # Override built-in defaults
     Defaults                syslog=auth
     Defaults>root           !set_logname
     Defaults:FULLTIMERS     !lecture
     Defaults:millert        !authenticate
     Defaults@SERVERS        log_year, logfile=/var/log/sudo.log
     Defaults!PAGERS         noexec
```

I.e if you set a default, you can *override* it for specific cases. The most common case I know of is sudoreplay which requires:

```
Defaults log_output
``
but you have to exempt it for 2 things:

```
Defaults!/usr/bin/sudoreplay !log_output
Defaults!/sbin/reboot !log_output
```

or your shit breaks.

I see your point that having to know the difference between `!`, `:`, etc. isn't great, but that's **also** what you'll see in the sudoers man page, so it seems sane to have that be 1-1.

jaymzh:
@jaininshah9  - I've rebased for, you - should make importit not die. CC @davide125 and @NaomiReeves .

jaymzh:
FYI @jaininshah9 / @davide125 , @NaomiReeves - I believe the tests are failing because in the ShipIt config, `fb_sudo` isn't listed, so it didn't get pushed out when it was landed internally.

jaininshah9:
Yeah, I have a diff up (in review state) to add fb_sudo added to ship it. It should be fixed soon.

davide125:
Should be fixed with #109

PR #61, opened 2019-09-11
chef: fix travis again

davide125:
Differential Revision: D17330723


jaymzh:
Can you please include some context as to why this is necessary? It's not obvious why this won't work on travis.

davide125:
Yeah, this is a WIP, as there's currently multiple issues breaking travis, so I'm going through a process of elimination. I'll put in actual details once I've got it working.

PR #62, opened 2019-09-17
fb_apt: Make security mirror configurable, fix security wholes

jaymzh:
* The security mirror was hard-coded which won't work in an
  organization without outbound access
* The key management used short keyIDs which are insecure

### DIFF
--- a/cookbooks/fb_apt/attributes/default.rb
+++ b/cookbooks/fb_apt/attributes/default.rb
@@ -18,19 +18,22 @@
 
 if node.debian?
   mirror = 'http://httpredir.debian.org/debian'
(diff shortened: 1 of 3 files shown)


davide125:
Looks like this broke travis for ubuntu: https://travis-ci.org/facebook/chef-cookbooks/jobs/586430694

jaymzh:
Yeah travis was broken before you merged it, I hadn't had a chance to look.

Looking at it now, I'm not clear what's going on, but I'll try to look today.

davide125:
It's just a typo in the mirror URL. I'm landing a fix now, and I also think I finally managed to fix travis for good (but we'll know for sure in a few minutes).

Issue #63, opened 2019-09-18
fb_apt doesn't take into account /etc/apt/trusted.gpg.d/


jaymzh:
Debian and ubuntu both allow keys in `/etc/apt/trusted.gpg.d`

Currently if a key listed in chef is in a file in that directory, which for Ubuntu 18.04 all of the older keys are, then Chef won't find them, it tries to add them, this "succeeds" because it's already there and it tries to do this every time.

To make matters worse, `apt-key` does not accept more than one `--keyring` option.

I'm not sure of the best solution, but I suspect it's:

* Chef should use a keyring like `/etc/apt/trusted.gpg.d/fb_keyring.gpg`
* Chef should, by default, cleanup keyrings that dont' match the format of `{ubuntu,debian}-keyring-\d{4}-.*.gpg` but should leave that and the main trusted.gpg alone as "the package's pervue"
* Users should then only specify non-OS keys in chef.

jaymzh:
current workaround, in my site_settings:

```
# https://github.com/facebook/chef-cookbooks/issues/63
node.default['fb_apt']['keyring'] = '/etc/apt/trusted.gpg.d/fb_keyring.gpg'
node.default['fb_apt']['keys'] = {}
```

jaymzh:
That didn't work... even with a fix I had to make, I had to make `keyring` be nil to avoid it touching keys at all.

jaymzh:
OK so it gets a bit worse:

* `fb_apt` isn't passing `--keyring #{keyring}` into `apt-key add`... but even if you do, it's totally ignored
* As such, specifying a key will always give you an asymmetric bug - so the only sane way to do things is to *never* specify a keyring. This means we will see all loaded keys, and never try to load a key that's loaded. But it ALSO means all keys we want to mess with are loaded into `/etc/apt/trusted.gpg`, which is probably fine.
* Debian has stopped putting *any* official keys in the main keyring - they are all in files in `/etc/apt/trusted.gpg.d/`, the keys I have on the main keyring are all from other repos I have - dropbox, bjn, etc.

So, I can see two paths forward:

# Option 1
* Drop the `node['fb_apt']['keyring']` option
[...]

PR #66, opened 2019-09-19
fb_apt: redo how key management works.

jaymzh:
[You can see https://github.com/facebook/chef-cookbooks/issues/63 for
the history here.]

This revamps the way key management works in fb_apt. It does the
following things:

* Fixes a bug where we deleted all keys because we weren't looking
at the `keys()` of the hash
* No longer tries to only look at a single keyring which does not
work reliably
* Pulls in all keyrings owned by packages and treats them as trustworthy
* Cleanups keyrings not owned by packages (configurable)
* Since we're now trusting keyrings from packages, this also validates
those files haven't been tampered with (configurable)

### DIFF
--- a/cookbooks/fb_apt/libraries/default.rb
+++ b/cookbooks/fb_apt/libraries/default.rb
@@ -43,5 +43,92 @@ module FB
         return "\n#{indent}#{k} \"#{v}\";"
       end
     end
+
+    # Grab all keyrings owned by a package. We do not include
+    def self._get_owned_keyring_files(node)
+      Dir.glob('/etc/apt/trusted.gpg.d/*').select do |keyring|
+        cmd = dpkg("-S #{keyring}")
+        # reject this file if it isn't in a package...
+        next false if cmd.error?
+
+        # if it is in a package, double check it's not been modified
+        pkg = cmd.stdout.lines.first.split(':').first
+        cmd = dpkg("-V #{pkg}")
+        modified_files = cmd.stdout.lines.map { |line| line.split.last }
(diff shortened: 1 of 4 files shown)


cookbooks/fb_apt/libraries/default.rb
```diff
@@ -43,5 +43,82 @@ def self._gen_apt_conf_entry(k, v, i = 0)
         return "\n#{indent}#{k} \"#{v}\";"
       end
     end
+
+    # Grab all keyrings owned by a package. We do not include
+    def self._get_owned_keyring_files
+      Dir.glob('/etc/apt/trusted.gpg.d/*').select do |keyring|
+        cmd = Mixlib::ShellOut.new("dpkg -S #{keyring}").run_command
```
davide125: Always pass LANG=C when calling debian tools, otherwise you may get output in whatever the random locale the system is using

cookbooks/fb_apt/libraries/default.rb
```diff
@@ -43,5 +43,82 @@ def self._gen_apt_conf_entry(k, v, i = 0)
...
+        # reject this file if it isn't in a package...
+        next false if cmd.error?
+
+        # if it is in a package, double check it's not been modified
+        pkg = cmd.stdout.lines.first.split(':').first
+        cmd = Mixlib::ShellOut.new("dpkg -V #{pkg}").run_command
```
davide125: Same here

jaymzh:
Updated, and abstracted out the running of commands so they're always run with `LANG=C`

PR #67, opened 2019-09-20
fb_storage: new cookbook to provision and manage devices and partitions

davide125:
Initial import of `fb_storage`, a new cookbook that allows one to manage
devices and partitions via an attribute-based API.

`fb_storage` is a rework and refactor of an internal implementation (the
"Chef Storage API") that was originally developed by @jaymzh in 2016, with
contributions of many others across the years.


cookbooks/fb_storage/libraries/storage_handlers.rb
```diff
@@ -0,0 +1,585 @@
...
+          self.mkfs_timeout = 900
+        end
+
+        # JBOD always matches as it'll work for every block device
+        def self.match?(_devname, _info)
+          true
+        end
+      end
+
+      class MdHandler < FB::Storage::Handler
```
jaymzh: I don't think the nvme handler is particularly private, did we drop that for some reason?

cookbooks/fb_storage/resources/format_devices.rb
```diff
@@ -0,0 +1,159 @@
...
+        ::File.read("/sys/block/#{dev}/queue/max_hw_sectors_kb").to_i
+    rescue StandardError
+      if ignore_failure
+        next
+      else
+        raise
+      end
+    end
+
+    if dev.start_with?('nvme')
```
jaymzh: ... speaking of nvme...

cookbooks/fb_storage/libraries/format_devices_provider.rb
```diff
@@ -0,0 +1,638 @@
...
+          # storage handler
+          sh = FB::Storage::Handler.get_handler(device, node)
+          sh.wipe_device
+          sh.prep_device
+          sh.partition_device(dev_settings)
+          sh.condition_device
+        end
+      end
+
+      def device_from_partition(partition)
```
jaymzh: this... is the same as `FB::Storage.device_name_from_partition` which is a class method you can call from here, so this method can be dropped and the calls to this in this file should call that.

cookbooks/fb_storage/libraries/storage.rb
```diff
@@ -0,0 +1,1120 @@
...
+      self._handle_custom_device_order_method(node)
+
+      prev = load_previous_disk_order
+
+      disk_to_slot_mapping = {}
+      if node['fb']['fbjbod']
+        unless node['fb']['fbjbod']['shelves'].keys.length.zero?
+          shelves = node['fb']['fbjbod']['shelves'].keys.sort
+          shelves.each do |shelf|
+            node['fb']['fbjbod']['shelves'][shelf].
```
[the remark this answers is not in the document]
davide125: Ah, good catch. Yeah, this should have been factored out from the start.

cookbooks/fb_storage/libraries/storage_handlers.rb
davide125: It currently shells out to the internal `flash_manager` to get card data. It will eventually get decoupled and opensourced as well, but it's going to take a while.

cookbooks/fb_storage/resources/format_devices.rb
davide125: Yeah, there's a bunch of enablement logic here and there that's currently only used by internal-only handlers. I figured it was still generic enough to leave around.

cookbooks/fb_storage/libraries/format_devices_provider.rb
davide125: Ah, yes, will refactor this.

davide125:
These have all been addressed, thanks @jaymzh!

PR #69, opened 2019-09-20
fb_dnsmasq: add a verifier

jaymzh:


cookbooks/fb_dnsmasq/recipes/default.rb
```diff
@@ -31,6 +31,7 @@
   owner 'root'
   group 'root'
   mode '0644'
+  verify 'dnsmasq --test -C %{path}'
```
joshuamiller01: Should this be an absolute path?

davide125: We don't really bother with fully qualified paths for verifiers in the rest of the code base, so let's get this landed for now.

PR #73, opened 2019-10-03
new cookbook: fb_sssd

jaymzh:
This is a very simple cookbook to manage `sssd` a common staple on
modern Linux machines - both servers and client machines.

I've tried to stick as close as possible to my recollection of the
internal FB cookbook. Other than some extra symlinks and a bit of
internal config, I think this should be a nearly drop-in replacement
that will be easy to migrate to.


cookbooks/fb_sssd/recipes/default.rb
```diff
@@ -0,0 +1,70 @@
...
+  action [:enable, :start]
+  subscribes :restart, 'template[/etc/nsswitch.conf]', :immediately
+end
+
+service 'disable sssd' do
+  not_if { node['fb_sssd']['enable'] }
+  # once the package is removed, this fails, sadly
+  only_if { File.exist?('/lib/systemd/system/sssd.service') }
+  service_name 'sssd'
+  action [:enable, :start]
```
davide125: You want `action [:stop, :disable]` here.

cookbooks/fb_sssd/recipes/default.rb
```diff
@@ -0,0 +1,70 @@
+#
+# Cookbook:: fb_sssd
+# Recipe:: default
+#
+# Copyright:: 2019, The Authors, All Rights Reserved.
```
davide125: pls add the usual boilerplate here (and in the attributes file)

davide125 (dismissed):

jaymzh:
Rebased. BTW, @mikedodge04 this one may be useful to you.

PR #74, opened 2019-10-03
An attempt to fix the rsyslog instability

jaymzh:
Test on GH are regularly failing because rsyslog fails to restart.
This is an attempt to fix that.

### DIFF
--- a/cookbooks/fb_syslog/recipes/default.rb
+++ b/cookbooks/fb_syslog/recipes/default.rb
@@ -77,4 +77,6 @@ end
 service service_name do
   action :start
   subscribes :restart, 'package[rsyslog]'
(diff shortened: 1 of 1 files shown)


jaymzh:
My fix works!

```
       Recipe: fb_syslog::default
         * service[rsyslog] action restart[2019-10-03T21:19:42+00:00] INFO: Processing service[rsyslog] action restart (fb_syslog::default line 77)
       [2019-10-03T21:19:42+00:00] INFO: Retrying execution of service[rsyslog], 0 attempt(s) left
       [2019-10-03T21:19:44+00:00] INFO: service[rsyslog] restarted
           - restart service service[rsyslog]
```

Some of these failed because of the PGP keyserver issues, but that's unrelated. I'll restart those and see if I can come up with a similar fix.

PR #75, opened 2019-10-03
Test all cookbooks

jaymzh:
Currently only cookbooks referenced in the sample init are tested,
but things like apache and friends that are likely to be included
from a "tier" role/cookbook and not from init are never tested.

This adds an entry point in testing to include those so they get
tested as well.

NOTE: This is based ON #77 so both changes are in it. Will rebase after that is landed (if it works)

### DIFF
--- a/cookbooks/fb_apt_cacher/recipes/default.rb
+++ b/cookbooks/fb_apt_cacher/recipes/default.rb
@@ -18,8 +18,8 @@
 # limitations under the License.
 #
 
-unless node.debian?
-  fail 'fb_apt_cacher is only supported on Debian.'
+unless node.debian? || node.ubuntu?
+  fail 'fb_apt_cacher is only supported on Debian-like Distros.'
 end
 
 package 'apt-cacher-ng' do
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -61,3 +61,4 @@ suites:
   - name: default
     run_list:
       - recipe[fb_init_sample]
+      - recipe[test_services]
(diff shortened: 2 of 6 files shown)


cookbooks/fb_apt_cacher/recipes/default.rb
```diff
@@ -18,7 +18,7 @@
 # limitations under the License.
 #
 
-unless node.debian?
+unless node.debian? || node.ubuntu?
   fail 'fb_apt_cacher is only supported on Debian.'
```
davide125: update the message as well

jaymzh:
Eh, I thought about that, but:
* Part of what I want to do here is converge on the sample_init being actually usable with minimal modifications. This gets us closer, but we're a long way from that.
* In order to exercise some of these "service" cookbooks in a sane way, we'll need to give them a sample configuration. Apache can start up with no vhosts, but we should pass it a vhost and make sure it starts up, for example. Throwing that in `init` definitely seems like the wrong place.

davide125:
Maybe we add an `fb_base_settings_sample` for the sample configuration? It'd double as an example of how to use this in the real world.

jaymzh:
We can do that, I know @mikedodge04 really wants to make a usable init, so I'd like to pull non-init things out to it's own cookbook. That cookbook is going to have weird disparate services in it, so it's (1) not reasonable to be used as-is and (2) could be used as a guide where applicable. I don't think that changes the argument that `init_sample` should not have bind and apache and nginx and whatever else in it. Especially since the configs there will have to ensure that apache, nginx and friends aren't trying to listen on the same port. :) Hence pulling this stuff to a separate testing-run-list cookbook seems better.

In addition to that, I think one thing this repo has lacked for a long time is an understanding of what the runlist might actually look like. It says it in the docs, but this actually shows "oh services go in a separate role/runlist" as we preach in the docs.

PR #80, opened 2019-10-04
fb_grub: lots of fixes

jaymzh:
* Modern debians just use `/boot/grub`, not `/boot/grub2`
* Don't redefine the directory, use the central definitions
* Allow people to not right both EFI and BIOS versions which breaks
on systems without EFI dirs
* Handle the `filesystem` and `filesystem2` situation (as of 14, FS2 is
no more, it's just FS)
* when grub and grub2's dirs are the same, don't create and then delete
them
* Update fb_helpers to also handle the `filesystem` v `filesystem2`
conversion

### DIFF
--- a/cookbooks/fb_grub/attributes/default.rb
+++ b/cookbooks/fb_grub/attributes/default.rb
@@ -17,10 +17,15 @@
 #
 
 version = node.centos6? ? 1 : 2
+grub2_base_dir = '/boot/grub2'
 if node.centos6? || node.redhat?
   vendor = 'redhat'
 elsif node.debian?
+  grub2_base_dir = '/boot/grub'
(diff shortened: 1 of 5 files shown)


cookbooks/fb_grub/attributes/default.rb
```diff
@@ -65,6 +70,7 @@
   'timeout' => 5,
   'use_labels' => false,
   'version' => version,
+  'force_write_both_efi_and_bios' => true,
```
davide125: This is referenced as `force_both_efi_and_bios` later on, so one of them is wrong. Also, you should add it to the README.

jaymzh:
Addressed review comments.

PR #82, opened 2019-10-18
fb_systemd: fix default_target confusion

jaymzh:
Currently there's no validation of the default target which can
render your system booting incorrectly. This fixes that.

### DIFF
--- a/cookbooks/fb_systemd/recipes/default.rb
+++ b/cookbooks/fb_systemd/recipes/default.rb
@@ -128,6 +128,13 @@ directory '/etc/systemd/user/default.target.wants' do
   mode '0755'
 end
 
-link '/etc/systemd/system/default.target' do
-  to lazy { node['fb_systemd']['default_target'] }
+execute 'set default target' do
+  not_if do
+    current = Mixlib::ShellOut.new('systemctl get-default').run_command.
+      stdout.strip
+    current != node['fb_systemd']['default_target']
+  end
+  command lazy {
+    "systemctl set-default #{node['fb_systemd']['default_target']}"
+  }
 end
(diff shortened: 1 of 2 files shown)


cookbooks/fb_systemd/recipes/default.rb
```diff
@@ -128,6 +128,13 @@
   mode '0755'
 end
 
-link '/etc/systemd/system/default.target' do
-  to lazy { node['fb_systemd']['default_target'] }
+execute 'set default target' do
+  not_if do
+    current = Mixlib::ShellOut.new('systemctl get-default').run_command.
+      stdout.strip
+    current != node['fb_systemd']['default_target']
```
joshuamiller01: Shouldn't we not do this when the current is the same as the default_target?  You have it backwards.

jaymzh:
oops, fixed. Thanks!

cookbooks/fb_systemd/recipes/default.rb
```diff
@@ -128,6 +128,13 @@
   mode '0755'
 end
 
-link '/etc/systemd/system/default.target' do
-  to lazy { node['fb_systemd']['default_target'] }
+execute 'set default target' do
+  only_if do
+    current = Mixlib::ShellOut.new('systemctl get-default').run_command.
```
joshuamiller01: Should we bail out here if this fails? i.e. `error!`

jaymzh: I don't believe it can fail, but even if it does, the set-default will fail because it validates the target is valid, which wills till bail out. 

That said I have no problems adding `fail ... if ...`

davide125: get-default goes over dbus (https://github.com/systemd/systemd/blob/master/src/systemctl/systemctl.c#L2132), so it could fail if dbus is hosed or if pid 1 is deadlocked

PR #86, opened 2019-10-28
Fix resolved handling: enable nss-resolve opt-out

jaymzh:
systemd releases before [this commit](https://github.com/systemd/systemd/pull/2031) had a non-functional `nss-resolve`. `systemd-resolved` works fine, but
not the nss plugin. Therefore binding them together doesn't work.

This provides the ability to turn off nss-resolve even when using
resolved.

### DIFF
--- a/cookbooks/fb_systemd/README.md
+++ b/cookbooks/fb_systemd/README.md
@@ -23,6 +23,7 @@ Attributes
 * node['fb_systemd']['logind']['config']
 * node['fb_systemd']['networkd']['enable']
 * node['fb_systemd']['resolved']['enable']
+* node['fb_systemd']['resolved']['enable_nss_resolve']
 * node['fb_systemd']['resolved']['config']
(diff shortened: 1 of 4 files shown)


cookbooks/fb_systemd/README.md
```diff
@@ -22,7 +22,8 @@ Attributes
 * node['fb_systemd']['logind']['enable']
 * node['fb_systemd']['logind']['config']
 * node['fb_systemd']['networkd']['enable']
-* node['fb_systemd']['resolved']['enable']
+* node['fb_systemd']['networkd']['enable']
+* node['fb_systemd']['resolved']['enable-nss-resolve']
```
davide125: I believe the convention is to use underscores instead of dashes for these things (i.e. `enable_nss_resolve`)

jaymzh:
Good call.

jaymzh:
Updated as requested.

PR #87, opened 2019-11-01
Fix disable idempotency issues with journald-*

jaymzh:
Due to https://github.com/chef/chef/issues/9041 `disable` doesn't work
on indirect units. There are only 2 so we'll use `systemd_unit` for
them.

Note that:

1. I'm ONLY changing the disable/stop units since there are
notifications against `service[$UNIT]` both in this and potentially
other cookbooks
2. Even though the resource name changing means I could drop the
explicit `unit_name`/`service_name` piece and keep the service/unit as
the resource name, doing so would break the pattern of these cookbooks,
so I didn't.

### DIFF
--- a/cookbooks/fb_systemd/recipes/journal-gatewayd.rb
+++ b/cookbooks/fb_systemd/recipes/journal-gatewayd.rb
@@ -27,6 +27,7 @@ end
 # https://github.com/chef/chef/issues/9041
 systemd_unit 'disable systemd-journal-gatewayd' do
   not_if { node['fb_systemd']['journal-gatewayd']['enable'] }
+  # ~FC009 - old version of FC...
   unit_name 'systemd-journal-gatewayd.service'
   action [:stop, :disable]
 end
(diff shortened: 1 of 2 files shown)


cookbooks/fb_systemd/recipes/journal-gatewayd.rb
```diff
@@ -23,8 +23,10 @@
   action [:enable, :start]
 end
 
-service 'disable systemd-journal-gatewayd' do
+# Need to use systemd_unit becuase of
```
davide125: s/becuase/because/

cookbooks/fb_systemd/recipes/journal-remote.rb
```diff
@@ -42,8 +42,10 @@
   action [:enable, :start]
 end
 
-service 'disable systemd-journal-remote' do
+# Need to use systemd_unit becuase of
```
davide125: s/becuase/because/

PR #89, opened 2019-11-08
fb_apt: significant speedup; os upgrades

jaymzh:
This PR changes `fb_apt` from running a `dpkg -S` on every single
keyring, and instead runs a single one. This is significantly faster,
saving ~3 seconds on a machine with only a few keyrings.

This also allows the user to override the distro which is necessary
for instrumenting OS upgrades across distros.

### DIFF
--- a/cookbooks/fb_apt/libraries/default.rb
+++ b/cookbooks/fb_apt/libraries/default.rb
@@ -66,7 +66,7 @@ module FB
         # keys that in both sets are modified keys
         modified_keys = owned_keys & modified_files
         Chef::Log.debug(
-          "fb_apt[keys]: Modofied keys from #{pkg}: #{modified_keys}",
+          "fb_apt[keys]: Modified keys from #{pkg}: #{modified_keys}",
         )
         unless modified_keys.empty?
           if node['fb_apt']['allow_modified_pkg_keyrings']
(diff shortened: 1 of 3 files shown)


jaymzh:
it's passing rubocop now

cookbooks/fb_apt/libraries/default.rb
```diff
@@ -46,31 +46,44 @@ def self._gen_apt_conf_entry(k, v, i = 0)
...
+      packages.each do |pkg|
         cmd = dpkg("-V #{pkg}")
-        modified_files = cmd.stdout.lines.map { |line| line.split.last }
-        if modified_files.include?(keyring)
+        modified_files =
+          Set.new(cmd.stdout.lines.map { |line| line.split.last })
+        # keys that in both sets are modified keys
+        modified_keys = owned_keys & modified_files
+        Chef::Log.debug(
+          "fb_apt[keys]: Modofied keys from #{pkg}: #{modified_keys}",
```
davide125: s/Modofied/Modified/

cookbooks/fb_apt/providers/sources_list.rb
```diff
@@ -25,7 +25,7 @@ def whyrun_supported?
 action :run do
   mirror = node['fb_apt']['mirror']
   security_mirror = node['fb_apt']['security_mirror']
-  distro = node['lsb']['codename']
+  distro = node['fb_apt']['distro'] || node['lsb']['codename']
```
davide125: Mind adding a comment here to clarify?

jaymzh: done. Also added it to README.md

cookbooks/fb_apt/libraries/default.rb
jaymzh: Done.

PR #90, opened 2019-11-22
fb_fstab: various fixes

jaymzh:
What follows is a variety of small fixes Vicarious has made to fb_fstab
while rolling it out. The Ohai one is from me, and the rest are from
@bwann.

* ohai-reload should not count as an "updated resource" here, and more
  importantly, shouldn't cause the entirety of fb_fstab to "update"
  every run, so we use `build_resource` to avoid that
* When `nfs4` is the filesystem, it should get all the same special
  "remote filesystem" treatment as `nfs`
* Whatever the `/` filesystem is needs to be considered 'core'
  filesystem. Obviously.
* Don't mount `noauto` filesystems
* Some better handling for systemd automounts

### DIFF
--- a/cookbooks/fb_fstab/libraries/default.rb
+++ b/cookbooks/fb_fstab/libraries/default.rb
@@ -25,6 +25,7 @@ module FB
       def self.determine_base_fstab_entries(full_fstab)
         core_fs_line_matching = [
           '^LABEL=(\/|\/boot|SWAP.*|\/mnt\/d\d+)\s',
+          '^\S+\s/\s',
           '^UUID=',
           '^devpts',
           '^sysfs',
--- a/cookbooks/fb_fstab/libraries/default.rb
+++ b/cookbooks/fb_fstab/libraries/default.rb
@@ -56,7 +56,7 @@ module FB
       end
 
       def self.generate_base_fstab
-        unless File.exist?(BASE_FILENAME) && File.size(BASE_FILENAME) > 0
+        unless File.exist?(BASE_FILENAME) && File.size?(BASE_FILENAME)
           FileUtils.cp('/etc/fstab', '/root/fstab.before_fb_fstab')
           FileUtils.chmod(0400, '/root/fstab.before_fb_fstab')
           full_fstab = File.read('/etc/fstab')
@@ -77,6 +77,7 @@ module FB
       # Returns an array of disks
       def self.get_in_maint_disks
         return [] unless File.exist?(IN_MAINT_DISKS_FILENAME)
+
         age = (
           Time.now - File.stat(IN_MAINT_DISKS_FILENAME).mtime
         ).to_i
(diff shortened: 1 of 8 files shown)


cookbooks/fb_fstab/libraries/default.rb
```diff
@@ -25,6 +25,7 @@ class Fstab
       def self.determine_base_fstab_entries(full_fstab)
         core_fs_line_matching = [
           '^LABEL=(\/|\/boot|SWAP.*|\/mnt\/d\d+)\s',
+          '^\S+\s/\s',
```
bwann: This matches " / " for those that miss it

bwann: specifically, something like `/dev/mapper/root / `

cookbooks/fb_fstab/libraries/provider.rb
```diff
@@ -596,6 +596,17 @@ def check_wanted_filesystems
           next
         end
 
+        if desired_data['opts']
+          opt_list = desired_data['opts'].split(',')
+          if opt_list.include?('noauto')
+            Chef::Log.debug(
+              "fb_fstab: '#{desired_data['device']}' is configured with " +
+              "'noauto' option, we will not mount."
```
NaomiReeves: lint
```suggestion
              "'noauto' option, we will not mount.",
```

cookbooks/fb_fstab/libraries/default.rb
```diff
@@ -205,26 +208,29 @@ def self.get_unmasked_base_mounts(format, node)
               data['mount_point'] == line_parts[1] &&
                 data['device'].start_with?('LABEL=')
             end
+
             raise e
           end
           # If someone has a more specific mount, don't use the original
           next if desired_mounts.any? do |_name, data|
             begin
```
jaymzh: This changes a bit because rubocop chokes on this crazy mix of things, trying to do an implicit begin and then breaking things, so this is both a bit cleaner, more clear, and doesn't confuse rubocop

cookbooks/fb_fstab/spec/public_spec.rb
```diff
@@ -58,7 +57,6 @@
 proc                    /proc                   proc    defaults        0 0
 tmpfs /dev/shm tmpfs defaults,size=4G 0 0
 EOF
-  # rubocop:enable LineLength
 
   context 'generate_base_fstab' do
     it 'should not regenerate base fstab' do
```
NaomiReeves: Please update this test, line 65 to reflect the change from `File.size` to `File.size?` in the method:
```
--- a/chef/cookbooks/core/fb_fstab/spec/public_spec.rb
+++ b/chef/cookbooks/core/fb_fstab/spec/public_spec.rb
@@ -62,7 +62,7 @@
     it 'should not regenerate base fstab' do
       File.should_receive(:exist?).with(FB::Fstab::BASE_FILENAME).
         and_return(true)
-      File.should_receive(:size).with(FB::Fstab::BASE_FILENAME).
+      File.should_receive(:size?).with(FB::Fstab::BASE_FILENAME).
         and_return(12)
       File.should_not_receive(:open)
       FB::Fstab.generate_base_fstab
```

PR #92, opened 2019-12-13
fb_systemd: networkd, resolved, and udev fixes

jaymzh:
* fb_systemd doesn't reload the right daemons when the packages are
upgraded
* fb_systemd lays down an udev rules file that is overridden by
everything else
* Don't claim you're updating resources when you're not

### DIFF
--- a/cookbooks/fb_systemd/recipes/networkd.rb
+++ b/cookbooks/fb_systemd/recipes/networkd.rb
@@ -23,6 +23,7 @@
 }.each do |svc|
   service svc do
     only_if { node['fb_systemd']['networkd']['enable'] }
+    subscribes :restart, 'package[systemd packages]', :immediately
     action [:enable, :start]
   end
 
--- a/cookbooks/fb_systemd/recipes/resolved.rb
+++ b/cookbooks/fb_systemd/recipes/resolved.rb
@@ -35,7 +35,7 @@ end
 # resolver and is required for systemd-resolved to work. This block attempts
 # to place the resolver between mymachines and myhostname as recommended by
 # upstream.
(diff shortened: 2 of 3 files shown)


cookbooks/fb_systemd/recipes/networkd.rb
```diff
@@ -23,6 +23,7 @@
 }.each do |svc|
   service svc do
     only_if { node['fb_systemd']['networkd']['enable'] }
+    subscribes :restart, 'package[systemd packages]', :immediately
```
davide125: Won't this cause issues if `manage_systemd_packages` is false? It might be better to do this as a notify instead

jaymzh: Why would it? that resource still exists (its not in an `if`, but never actually fires `it has an `only_if`. This should work fine. However, I certainly can hang all of these off the package, I was following suite since you have no notifies on that package... and since you subscribe to it from `boot.rb`, I was following suite. Since someone might carve these things up and not include specific recipes and thus not *have* these various services, the `notifies` is a bit more brittle, I think. `subscribes` allows the thing it's subscribing to to not exist.

davide125: Yeah, I think you're right actually.

davide125:
I'll get this imported. Might take a while to merge as I suspect I'll have to shard the udev change to avoid reloading it everywhere at once.

jaymzh:
@davide125 wanted to follow up here...

davide125:
Yeah, I'm testing this out now, we have a few places messing with udev that require reshuffling before we can merge this.

PR #94, opened 2019-12-20
Some more docs for chef-cookbooks

jaymzh:
This is an attempt to take a lot of the concepts, rules, and guidelines
that are documented internally at FB and write them for external
consumption.

### DIFF
--- a/README.md
+++ b/README.md
@@ -241,7 +241,7 @@ that's more convient to use in your template.
 ### Implementating runtime-safe APIs
 
 This model intentionally draws the complexity of Chef into the "core cookbooks"
-(those implementing APIs) so tha the user experience of maintaining systems is
+(those implementing APIs) so that the user experience of maintaining systems is
 simple and (usually) requires little more than writing to the node object.
 However, the trade-off for that simplicity is that implementing the API properly
 can be quite tricky.
--- a/README.md
+++ b/README.md
@@ -138,26 +138,39 @@ and then create a class for their cookbook under that. For example, `fb_fstab`
 creates a `class Fstab` inside of the `module FB`. We will refer to this as
 the cookbook class from here.
 
-We require all cookbooks use this model for consitency.
+We require all cookbooks use this model for consistency.
 
 Since we don't put anything other than other classes inside the top-level
 object, it's clear that a `module` is the right choice.
 
-While there is no reason that a cookbook class can't be one design to be
-instantiated, more often than not it is simply a collection of callable class
-methods - a place to put methods that can then be called both from this
-cookbook and others.
+While there is no reason that a cookbook class can't be one designed to be
+instantiated, more often than not it is simply a collection of class methods and
+constants (i.e. static data and methods that can then be called both from this
+cookbook and others).
 
 Below the cookbook class, the author is free to make whatever class or methods
 they desire.
 
-It is often desirable to `include` a module into a Custom Resource, and we do
-this by creating a module inside of the cookbook class, often ending in
-`Provider`, e.g. `FstabProvider`. This code should be in its own library file.
+When building a complicated Custom Resource, the recommended pattern is to
+factor out the majority of the logic into a module, inside of the cookbook
+class, that can be `include`d in the `action_class`. This allows the logic to be
+easily unit tested using simple rspec. It is preferred for this module to be in
+its own library file, and for its name to end in `Provider`, ala
+`FB::Thing::ThingProvider`.
 
-It is worth noting that `fb_fstab` itself violates this guideline as it is
-quite old before we had solidified a standard and thus it's provider module
-is directly below the `module FB`.
+When more than 1 or 2 methods from this module are called from the custom
+resource itself, it is highly recommended you include it in a Helper class for
+clarity, ala:
+
+```
+action_class do
+  class ThingHelper
+    include FB::Thing::ThingProvider
+  end
+end
+```
+
+In this way, it is clear where methods come from.
 
 ### Extending the node vs self-contained classes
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -111,6 +124,268 @@ instructions in its
...
+It is often useful to factor out logic into a library - especially logic that
+doesn't create resources. Doing so makes this logic easier to unit test and
+makes the recipe or resource cleaner.
+
+Our standard is that all cookbooks use the top-level container of `module FB`,
+and then create a class for their cookbook under that. For example, `fb_fstab`
+creates a `class Fstab` inside of the `module FB`. We will refer to this as
+the cookbook class from here.
+
+We require all cookbooks use this model for consitency.
```
davide125: s/consitency/consistency/

README.md
```diff
@@ -111,6 +124,268 @@ instructions in its
...
+
+Since we don't put anything other than other classes inside the top-level
+object, it's clear that a `module` is the right choice.
+
+While there is no reason that a cookbook class can't be one design to be
```
davide125: I think you meant "designed", not "design" here

README.md
```diff
@@ -111,6 +124,268 @@ instructions in its
...
+instantiated, more often than not it is simply a collection of callable class
+methods - a place to put methods that can then be called both from this
+cookbook and others.
```
davide125: Another common usecase worth mentioning is static data: constants, enums, etc

README.md
```diff
@@ -111,6 +124,268 @@ instructions in its
...
+
+Below the cookbook class, the author is free to make whatever class or methods
+they desire.
+
+It is often desirable to `include` a module into a Custom Resource, and we do
```
davide125: I don't know if I'd say this is "often" desirable -- while it's definitely ok if it's done for a reason, I find in many cases it makes things needlessly more complex.

README.md
```diff
@@ -111,6 +124,268 @@ instructions in its
...
+this by creating a module inside of the cookbook class, often ending in
+`Provider`, e.g. `FstabProvider`. This code should be in its own library file.
+
+It is worth noting that `fb_fstab` itself violates this guideline as it is
+quite old before we had solidified a standard and thus it's provider module
```
davide125: s/it's/its/. Also maybe rewrite this whole sentence to read less like a stream of consciousness

README.md
```diff
@@ -111,6 +124,268 @@ instructions in its
...
+is directly below the `module FB`.
+
+### Extending the node vs self-contained classes
+
+You may have noticed that some of our cookbooks will extend the `node` object,
+while others have self-contained classes that sometimes require the `node` be
+passed as a parameter to some methods.
+
+In general, the **only** time when extending the `node` is acceptable is when
+you are simply making a convenience function around using the node object. So,
```
davide125: I agree with this, and we both know what you mean here, but "convenience function" isn't much of a guideline for someone that's not already familiar with this stuff. Can/should they consume ohai data? node data from the cookbook? from other cookbooks? Can they do if then else logic or should it just be one line? Note that we currently have all of the above in the codebase.

README.md
```diff
@@ -111,6 +124,268 @@ instructions in its
...
+for example, instead of making people do `node['platform_family'] = 'debian'`,
+there's a `node.debian?`. This is simply syntactic sugar ontop of data entirely
```
davide125: s/ontop/on top/

README.md
```diff
@@ -111,6 +124,268 @@ instructions in its
...
+in the node.
+
+In all other cases, one should simply have the `node` be an arguement passed on,
+so as to not pollute the node namespace. For example, a method that looks at the
+node attributes, but also does a variety of other logic, should be in a cookbook
+class and take the node as an argument (per standard programming paradigms about
+clear dependencies).
+
+### Methods in recipes
```
davide125: I'd also expand this to cover methods (and other complex logic) in templates (or maybe add a separate section for that)

README.md
jaymzh: Are you responding to moving the logic into a module  or to `include`ing the module?

WRT to the first, I **strongly** disagree. For anything short of the simplest providers, it's the only way to make it reasonably unit-tested. Otherwise you end up mocking a ton of stuff in chefspec... and while I likes chefspec, it's far far far far easier to rspec-test a class than it is to chefspec test a provider, *especially* a provider in our model which is an "internal" provider.

Starting by sticking your methods in a class is far far simpler and will lead to cleaner code.

Now, wrt to the `include`ing, it gets a bit fuzzier. Bty default is' the same as python's multiinheritence, but that also is a bit implicit, so what I've been doing recently is:

```
action_class do
  class FbThingHelper
    include FB::Thing::Provider
  end
end
```

So that at least then all the methods are `FbThingHelper.whatever()` so it's at least clear where they come from.

jaymzh:
all your other comments are :+1:  - I'll address them. Just that one, I (possibly) disagree with.

README.md
jaymzh: This is why I included some examples. I think it's impossible to provide a hard line, this is mostly going to be up to our judgement on PR reviews, but this is a generally useful starting point, I think.

README.md
jaymzh: I rewrote this...

README.md
jaymzh: This paragraph is gone.

jaymzh:
OK updated, I think that addresses everything.

README.md
davide125: Yeah, my concern was specifically about the `include`'ing, which I've found confusing to explain to people in the past. I like the usage of a helper class like this though, I think it makes the separation clearer and at least avoid polluting the namespace.

PR #98, opened 2020-01-09
Fix device_name_from_partition

jaymzh:
Use the right algorithm to determine when there is a 'p'

### DIFF
--- a/cookbooks/fb_storage/libraries/storage.rb
+++ b/cookbooks/fb_storage/libraries/storage.rb
@@ -110,8 +110,27 @@ module FB
       "#{device}#{prefix}#{partnum}"
     end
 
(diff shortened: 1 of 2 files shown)


jaymzh:
@NaomiReeves very, very astutely found that @davide125 did this before in https://github.com/facebook/chef-cookbooks/commit/3b1ea4664f6d1416df84d3d1b617e39970780bf4#diff-591f238ac592359f57cb17dd8bea076d Igor had to somewhat revert this in https://github.com/facebook/chef-cookbooks/commit/22d564a3be86a5258c4a404da997bfc3901a3fe2 - which is a bummer.

But the code restored wasn't quite right either. :) This new code handles both cases and adds a hefty comment. test coming too

PR #102, opened 2020-02-18
Move fb_fstab to use fb_helpers

jaymzh:
This depends on #101 which moves this method to a more accessable
location, in `fb_helpers`.


jaymzh:
(this will fail until #101 is merged and I rebase this)

PR #103, opened 2020-02-18
fb_logrotate: debian compatibility

jaymzh:
Credit here goes to @bwann who I believe actually made these changes,
I'm just doing the syncing.


jaymzh:
I think you're misreading. They're just temp variables at the top, the right one gets assigned to `configs` just like in every other cookbook :) See line 86

PR #105, opened 2020-02-18
New cookbook: fb_profile

jaymzh:
A cookbook to manage global profile files using the FB attribute-driven
model.

### DIFF
--- a/cookbooks/fb_init_sample/metadata.rb
+++ b/cookbooks/fb_init_sample/metadata.rb
@@ -44,6 +44,7 @@ end
   'fb_motd',
   'fb_nsswitch',
   'fb_postfix',
+  'fb_profile',
   'fb_rpm',
   'fb_rsync',
   'fb_sdparm',
--- a/cookbooks/fb_init_sample/recipes/default.rb
+++ b/cookbooks/fb_init_sample/recipes/default.rb
@@ -99,6 +99,7 @@ if node.centos? && !node.container?
   include_recipe 'fb_ipset::cleanup'
 end
 include_recipe 'fb_motd'
+include_recipe 'fb_profile'
 
 if node.firstboot_tier?
   include_recipe 'fb_init_sample::firstboot'
--- a/cookbooks/fb_profile/README.md
+++ b/cookbooks/fb_profile/README.md
@@ -1,0 +1,48 @@
+fb_profile Cookbook
+===================
+
+Requirements
+------------
+
+Attributes
+----------
+* node['fb_profile']['early_entries']
+* node['fb_profile']['aliases']
+* node['fb_profile']['variables']
+* node['fb_profile']['late_entries']
+
+Usage
+-----
+This generates a file in `/etc/profile.d`, which means that what
+you put in here should be bourne-shell compatible. It also means
+it's only loaded once at login, not on each terminal you open.
+
+### Aliases
+
+This is a simple hash of alias to command. Ala:
+
+```ruby
+node.default['fb_profile']['aliases']['ll'] = 'ls -l'
+```
+
+### Variables
+
+This is a straight forward mapping of variables:
+
+```ruby
+node.default['fb_profile']['variables']['EDITOR'] = 'vim'
+```
+
+All variables are `export`ed.
+
+### Early and Late entries
+These are wholesale entries that are put at the beginning or end of the file.
+It is recommended you used heredocs with them like so:
+
+```ruby
+node.default['fb_profile']['early_entries']['my cool thing'] = << ~EOF
+  if [ "$FOO" = 'yes' ]; then
+    # do a thing here
+  fi
+EOF
+```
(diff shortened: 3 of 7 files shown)


davide125:
Funnily enough, we have an internal `fb_etc_profiled` targeting Fedora that we were about to publish. I'll see if I can get it merged with this one instead, as they look similar enough. cc @michel-slm

jaymzh:
This one is incredibly simple... if yours is better, I'm happy to migrate to it. I just needed something. I was planning on adding an overwrite option to /etc/bash.bashrc which would include a file as well that we could maintain since it lacks such a thing...

davide125:
@jaymzh our internal version is at https://github.com/davide125/chef-cookbooks/tree/fb_etc_profiled/cookbooks/fb_etc_profiled

I'm going to hold off on merging it in this repo, as I'd rather get a unified solution. @michel-slm mentioned one of the requirements for us is that this will need to support OSX.

jaymzh:
Last I checked, OSX had dropped all support for global profile files...?

michel-slm:
Checking with our resident Mac expert on the macOS situation. We do know that we currently run fb_etc_profiled on Macs but if this is going away long-term I'm fine with not adding macOS support

jaymzh:
So I took a look over `fb_etc_profile_d`, to get an idea of which one I think should be merged into which one (I'm always happy to throw mine away if there's a more complete one).

In this case I think we probably want to base it on mine. Here's my thoughts on `fb_etc_profiled`:

1. I wouldn't override `/etc/profile` in the general sense - letting the OS do it's defaults here is probably best. It's a little non-standard for us to not just own an entire file let there be no other file, but in this case this isn't a finite-set of things. More below.
2. Dropping off multiple files in `/etc/profile.d` is pretty anti-fb-model here. The more fb-y thing to do would be to have a single file in there that implements whatever people pass into the API. Its just the same as we don't drop off `/etc/sysctl.conf.d/` files, we just manage `/etc/sysctl.conf` and (though in this case, there's not a full-known-set of things, hence I wouldn't overwrite `/etc/profile` itself). Here's a few specific reasons:
* We then need to clean up those files. This is do-able of course, but why add more work?
* We have to worry about source-ing order. We could, again, do some interesting work-arounds here - hashes in ruby are ordered so you could try to name them in the order you got them, but that leads to really un-fun stuff like "oh someone added a new entry so file after it got renamed!"
* It becomes much more difficult to see what changed in a given run because you have to piece together multiple diffs
(That's why in mine I make a `/etc/profile.d/fb_profile.sh`.)
3. There's no real API here - we're just jamming in entire files into variables via heredocs into the node object. It's hard to introspect, it's hard to code-review, it's hard to indent. Since, ultimately we'll generate a shell script, that functionality is of course necessary, for the common stuff, creating a nice API will help (ala [aliases](https://github.com/facebook/chef-cookbooks/pull/105/files#diff-b49441bfb7e32a6b9f2c9e85af74b4e7R25) and [variables](https://github.com/facebook/chef-cookbooks/pull/105/files#diff-b49441bfb7e32a6b9f2c9e85af74b4e7R33). This leads to some problems like:
* That model loses all the 'cascading' nature of Chef FB attribute model. We have *no* way of knowing who set an aliase for `ll`, nor do I have an easy way of overriding it. I just have to make my own and hope my files is loaded later. It also means, as you currently have it, If I call my "key" `zzz_fuck_everyone`, I win, even if I'm the first cookbook, rather than that runlist-order winning (though of course a later runlist could look for that and change it, but it's no longer using the intuitive order that is common in the FB model).

Now, all that said, mine is missing some useful features too and needs to grow, though I *think* as it stands it can cover for what this does.

Onto the big caveat. I said I would not overwrite `/etc/profile`... and in fact we should really never do that because every single distro's `/etc/profile` sources `/etc/profile.d/*` at the end (which means we probably should cleanup other stuff in that directory and just write our own single file, but that's already covered above, I think). **HOWEVER** I also want to implement the bashrc/zshrc files for non-login shells. When it comes to `/etc/bash.bashrc` (debian/ubuntu) or `/etc/bashrc` (fedora/rhel/centos) only the RH ones have a nice include where they re-source stuff in profile.d when necessary. This is great in that it makes it easy to not have to touch that file on RH-like OSes, but we'll probably *need* to on Debian-like OSes to start managing bashrc/zshrc globally, despite the fact that that's generally ill-advised.

Fortunately the debian/ubuntu `/etc/bash.bashrc` is only 23 lines, and doesn't change much, so overwriting it is... not the most-terribly-heinous thing ever because we can make ours those 23 lines plus the few lines we wanna "steal" from RH. :)

I hope that helps

michel-slm:
Given on macOS we can't rely on /etc/profile being loaded anyway, I'm fine with landing this as it is now, and then we can see internally if we can switch our Mac users to this (and finding a way to consistently load fb_profile.sh).

Right now fb_etc_profile_d only works on a Mac if you do a `bash --login`, not sure how it even works for the person who authored the call site that uses it.

PR #108, opened 2020-02-18
fb_systemd: 2 minor fixes

jaymzh:
* on firstboot create_tmpfiles can fail if it had to ignore things,
  so allow that return code
* handle networkd restart properly

### DIFF
--- a/cookbooks/fb_systemd/recipes/default.rb
+++ b/cookbooks/fb_systemd/recipes/default.rb
@@ -93,6 +93,11 @@ execute 'process tmpfiles' do
       map { |x| " --exclude-prefix=#{x}" }.
       join
   }
+  # it returns 65 if it had to ignore some lines, which seems to happen
+  # quite often on initial setup
+  if node.firstboot_any_phase?
+    returns [0, 65]
+  end
   action :nothing
 end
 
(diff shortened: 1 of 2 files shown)


cookbooks/fb_systemd/recipes/networkd.rb
```diff
@@ -24,6 +24,10 @@
   service svc do
     only_if { node['fb_systemd']['networkd']['enable'] }
     subscribes :restart, 'package[systemd packages]', :immediately
+    if svc == 'systemd-networkd.socket'
+      notifies :stop, 'service[systemd-networkd.service]', :before
+      notifies :start, 'service[systemd-networkd.service]', :immediately
+    end
```
davide125: Ah, thanks! We hit this internally as well on Ubuntu, and I had an in-progress diff to skip the socket here, but your approach is much better.

cookbooks/fb_systemd/recipes/default.rb
```diff
@@ -93,6 +93,11 @@
       map { |x| " --exclude-prefix=#{x}" }.
       join
   }
+  # it returns 65 if it had to ignore some lines, which seems to happen
+  # quite often on initial setup
+  if node.firstboot_any_phase?
+    returns [0, 65]
+  end
```
davide125: Should we warn something if it's skipping lines?

jaymzh: there's no good way to behave differently on different return values unless we want to write our own resource. In this case, since it's only on firstboot, I figure it's not worth it.

PR #113, opened 2020-03-19
stunnel cookbook - fix for c7

davide125:
Summary:

there's not systemd config in the c7 rpm, and we should use the standard
name for the config file.

Differential Revision: D20542181


davide125:
@jaymzh -- for context, it looks like on C7/C8 we don't read `/etc/sysconfig/stunnel` at all, so it never picks up the right parameters, including the config file

jaymzh:
OK it turns out it's a bit more complicated than that.

On Debian and Ubuntu, the service file is generated from the initscript and the initscript as actually quite complicated.

It *iterates* over every file in `FILES` (from the sysconfig / default file) and starts a separate instance for each one of those. It has a lot of logic to figure out if there's a chroot being used in each one so it can specify the correct pid file name and location. The default for FILES in the package is:
```
FILES="/etc/stunnel/*.conf"
```

We semi-explicitly in this cookbook "disable" that by making files a single value. There is no default stunnel.conf in that directory in Ubuntu, and so I picked `fb_stunnel.conf` to match `fb_apache` and friends.

Anyway - the simplest change, instead of changing the name is just update the unit file you are dropping off to append `/etc/stunnel/fb_stunnel.conf` to the end. The last argument to stunnel is a config file.

The better (IMO) change is to
* Set `/etc/sysconfig/stunnel` to an `EnvironmentFile` in your unit file and then just update your unit file to specify `$OPTIONS $FILES` - and then people can use `node['fb_stunnel']['sysconfig']['options']` to set command-line options and you've built in the proper functionality that debian is getting.
* Update the cookbook to *not* use `FILES` in the sysconfig file, skip it, and just hardcode `FILES="/etc/stunnel/fb_stunnel.conf"` in the template since that's the file we generate

This will:
* Be more clear
* Make both OSes behavior fairly similarly
* Provide people the ability set `options`

PR #119, opened 2020-04-16
fb_chrony: Add debian/ubuntu support

jaymzh:
Debian/ubuntu use a different user for chrony

### DIFF
--- a/cookbooks/fb_chrony/recipes/default.rb
+++ b/cookbooks/fb_chrony/recipes/default.rb
@@ -22,9 +22,13 @@
 if node.centos? || node.redhat? || node.fedora?
   chrony_svc = 'chronyd'
   chrony_conf = '/etc/chrony.conf'
+  chrony_user = 'chrony'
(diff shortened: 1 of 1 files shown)


cookbooks/fb_chrony/recipes/default.rb
```diff
@@ -22,18 +22,22 @@
 if node.centos? || node.redhat? || node.fedora?
   chrony_svc = 'chronyd'
   chrony_conf = '/etc/chrony.conf'
+  chrony_user = 'chrony'
+  chrony_group = 'chrony'
 elsif node.debian_family?
   chrony_svc = 'chrony'
   chrony_conf = '/etc/chrony/chrony.conf'
+  chrony_user = '_chrony'
+  chrony_group = '_chrony'
```
davide125, line 31: Is this really `_chrony` (with the underscore) ?

jaymzh: Yes. :(

PR #120, opened 2020-04-16
fb_iptables: Fix dupe rules; output, and de-templatize

jaymzh:
**Summary**:

The primary purpose of this PR was to fix the fact that reloading meant
some dynamic chains ended up with duplicates of their rules which
quickly spun out of control as the ruleset was duplicated and duplicated
over and over. In order to do that I ended up cleaning up a bunch of
stuff that had gone awry with this code since I last touched it.

* Fix the bug in question - add a space (" ") after the name of the
  chain in the grep so that we don't end up grabbing all the rules for
  FOO-BAR when we're looking for FOO. Or in this case, DOCKER-USER would
  get caught by DOCKER and thus be duplicated.

* The `fb_iptables_reload` script wasn't supposed to be a template, it
  was erroneously made into a template in https://github.com/facebook/chef-cookbooks/commit/80a8a3a5e2fbe8f603230c2aabc1abba06d73c57#diff-80bee50f4cf6d0af9c03f58ea0f9cd44
  and this undoes that. The script **has a templatized config file**!!
  @davide125 should great shame for approving that PR. =P
* Fix the bug where we don't clear the chain before we load it causing
  duplicate rules
* Make the output actually make sense
* Fix the bug where if one dynamic chain on a table wasn't populated, we
  skipped restoring any of them

**Test Plan**:

The output looks like:
```
root@host:/etc/default# /usr/sbin/fb_iptables_reload 4 reload
Reloading mangle...
  - Stashing dynamic chains on mangle: done.
  - Reloading mangle
  - Restoring dynamic chains on mangle: done.
Reloading filter...
  - Stashing dynamic chains on filter: DOCKER-USER  DOCKER-ISOLATION-STAGE-1  DOCKER  DOCKER-ISOLATION-STAGE-2  done.
  - Reloading filter
  - Restoring dynamic chains on filter: DOCKER-USER DOCKER-ISOLATION-STAGE-1 DOCKER DOCKER-ISOLATION-STAGE-2 done.
Reloading raw...
  - Stashing dynamic chains on raw: done.
  - Reloading raw
  - Restoring dynamic chains on raw: done.
```

* The rules are no longer duplicated

* The config file is populated properly:

```
root@host:/etc/default# cat /etc/fb_iptables.conf
FILTER_CHAINS="DOCKER-USER DOCKER-ISOLATION-STAGE-1 DOCKER DOCKER-ISOLATION-STAGE-2"
MANGLE_CHAINS=""
RAW_CHAINS=""
CONFIG_DIR="/etc/iptables"
RULES_FILE="rules.v4"
RULES6_FILE="rules.v6"
TABLES="mangle filter raw"
STATIC_MANGLE_CHAINS=""
STATIC_FILTER_CHAINS="INPUT FORWARD"
STATIC_RAW_CHAINS=""
```

### DIFF
--- a/cookbooks/fb_iptables/files/default/fb_iptables_reload.sh
+++ b/cookbooks/fb_iptables/files/default/fb_iptables_reload.sh
@@ -1,4 +1,5 @@
 #!/bin/bash
+# shellcheck disable=SC2086,SC2018,SC2019,SC1090,SC2064,SC2124,SC2181,SC2002,SC2153,SC2046,SC2173,SC2230
 #
 # Copyright (c) 2016-present, Facebook, Inc.
 # All rights reserved.
@@ -42,15 +43,16 @@ dump_dynamic_chains() {
   local dynamic_chains="$2"
   local outfile
   local chain_regex
+  echo -n "  - Stashing dynamic chains on $table: "
   for chain in $dynamic_chains; do
-    echo -n "Stashing $chain on $table: "
+    echo -n "$chain  "
     "${IPTABLES_CMD}" -t "$table" -S | grep -q "\-N $chain"
     if [ $? -ne 0 ]; then
-      echo 'no chain yet, nothing to stash.'
-      return
+      echo '(no chain yet, nothing to stash) '
+      continue
     fi
     outfile=$TMPDIR/${table}_${chain}_rules
-    chain_regex="\-A $chain"
+    chain_regex="\-A $chain "
     out=$("${IPTABLES_CMD}" -t "$table" -S)
     if [ $? -ne 0 ]; then
       exit_error "Failed to stash $chain on $table"
@@ -58,8 +60,8 @@ dump_dynamic_chains() {
     if [ -n "$out" ]; then
       echo "$out" | grep "$chain_regex" > "$outfile"
     fi
-    echo 'done.'
   done
+  echo 'done.'
 }
 
 # Restore any registered chains we found and backed up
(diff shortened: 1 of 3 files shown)


cookbooks/fb_iptables/files/default/fb_iptables_reload.sh
```diff
@@ -1,4 +1,5 @@
 #!/bin/bash
+# shellcheck disable=SC2086,SC2018,SC2019,SC1090,SC2064,SC2124,SC2181,SC2002,SC2153,SC2046,SC2173
```
jaymzh, line 2: FYI - this is because this thing fails shellcheck badly and so I had to either lint it, or disable all this to merge it internally. Then I figured you wuoldn't want all of that change in this PR, so I'm first submitting the fixes... then I will submit fugure PRs to lint this.

davide125: Yeah that's fine. Add a TODO or a comment so we don't forget though.

davide125: Also, add SC2230 to the list, it's failing that too internally

jaymzh: That IS the TODO :) Also, it has to be on the 2nd line, so I can add another line below it saying fix, but I think the disabling all of lint is an obvious TODO. I also have an internal task to fix it
Yeah I can add that rule... let me know how testing goes in case I need to make any other changes.

davide125: Looking good so far, I'll merge it once you update this.

jaymzh:
OK @davide125 - I rebased and added that additional linter exclusion

Issue #123, opened 2020-05-01
fb_systemd::udevd fails on Ubuntu 20.04 LTS due to missing symlink


ChefAustin:
On Ubuntu 16 and 18 the `fb_systemd::udevd` recipe succeeds without issue but on Ubuntu 20 the recipe will fail because the symlink from `/sbin/udevadm` to `/bin/udevadm` has been removed.

From the `udev` filelists provided by Canonical for each release, one can validate the existence of `/sbin/udevadm` in [Xenial](https://packages.ubuntu.com/xenial/amd64/udev/filelist) and [Bionic Beaver](https://packages.ubuntu.com/bionic/amd64/udev/filelist), as well as its non-existence on [Focal Fossa](https://packages.ubuntu.com/focal/amd64/udev/filelist).

malmond77:
The PR is merged, thanks for the contribution!

PR #124, opened 2020-05-01
Add /sbin/udevadm > /bin/udevadm symlink before using sbin path

ChefAustin:
This pull request, if merged, will change `fb_systemd::udevd` ever-so-slightly by adding a symlink from `/sbin/udevadm` to `/bin/udevadm` at the beginning of the recipe.

The symlink creation is needed to mitigate an issue (#123) that is displayed when this recipe is run on Ubuntu 20 whereby the recipe fails due to Canonical's removal of the `/sbin/udevadm` symlink. By re-adding the symlink back to the system at the beginning of this recipe, it should ensure that the subsequent `trigger udev`, `reload udev`, and `update hwdb` `execute` resources proceed without path existence errors.

I considered a couple different alternative approaches to this fix but ultimately thought this was the most sane means of fixing this. Please feel free to dissent if you feel this approach is unsafe or not the best way to go about this.

### DIFF
--- a/cookbooks/fb_systemd/recipes/udevd.rb
+++ b/cookbooks/fb_systemd/recipes/udevd.rb
@@ -18,6 +18,12 @@
 # limitations under the License.
 #
 
+link '/sbin/udevadm' do
+  to '/bin/udevadm'
+  owner 'root'
+  group 'root'
+end
+
 execute 'trigger udev' do
   command '/sbin/udevadm trigger'
   action :nothing
--- a/cookbooks/fb_systemd/recipes/udevd.rb
+++ b/cookbooks/fb_systemd/recipes/udevd.rb
@@ -18,19 +18,26 @@
 # limitations under the License.
 #
 
+udevadm = value_for_platform(
+  'centos' => {
+    '< 6.0' => '/sbin/udevadm'
+  },
+  default => '/bin/udevadm',
+)
+
 execute 'trigger udev' do
-  command '/sbin/udevadm trigger'
+  command "#{udevadm} trigger"
   action :nothing
 end
 
 execute 'reload udev' do
-  command '/sbin/udevadm control --reload'
+  command "#{udevadm} control --reload"
   action :nothing
   notifies :run, 'execute[trigger udev]', :immediately
 end
 
 execute 'update hwdb' do
-  command '/sbin/udevadm hwdb --update'
+  command "#{udevadm} hwdb --update"
   action :nothing
 end
 
--- a/cookbooks/fb_systemd/recipes/udevd.rb
+++ b/cookbooks/fb_systemd/recipes/udevd.rb
@@ -20,9 +20,9 @@
 
 udevadm = value_for_platform(
   'centos' => {
-    '< 6.0' => '/sbin/udevadm'
+    '< 6.0' => '/sbin/udevadm',
   },
-  default => '/bin/udevadm',
+  'default' => '/bin/udevadm',
 )
 
 execute 'trigger udev' do
--- a/cookbooks/fb_systemd/recipes/udevd.rb
+++ b/cookbooks/fb_systemd/recipes/udevd.rb
@@ -18,12 +18,6 @@
 # limitations under the License.
 #
 
-link '/sbin/udevadm' do
-  to '/bin/udevadm'
-  owner 'root'
-  group 'root'
-end
-
 execute 'trigger udev' do
   command '/sbin/udevadm trigger'
   action :nothing


cookbooks/fb_systemd/recipes/udevd.rb
```diff
@@ -18,6 +18,12 @@
 # limitations under the License.
 #
 
+link '/sbin/udevadm' do
+  to '/bin/udevadm'
+  owner 'root'
+  group 'root'
+end
```
malmond77, line 25: Recommended changes:
1. Reverse the symlink
2. Use Relative paths so the link is meaningful outside of chroots/containers. `/bin/udevadm -> ../sbin/udevadm`
3. Adding a guard to avoid making the link if the target is a symlink (which is the normal case): `not_if { File.symlink?('/sbin/udevadm') }`
4. Change the paths below to use `/bin` directly.

davide125: @malmond77, it's not quite as simple, on modern Fedora and CentOS the whole of `/sbin` is a symlink to `/bin`, not just the one file. It might be easier to just make this path conditional using `value_for_platform` or similar instead of mucking with symlinks.

jaymzh: I second @davide125 's solution - though with a slight correction. It's not that `/sbin` and `/bin` are unified it's that `/bin`/`/usr/bin` have been unified and `/sbin`/`/usr/sbin` have been unified.

You can't reverse the symlink - as the OP points out the actual binary is in /bin - so you'd have to move the binary which seems like a bad idea. `value_for_platform` is definitely the right call here.

One other thing I'll add here is that *most* platforms have moved udevadm to bin at this point, so we should almost always be using /bin/udevadm

* Debian sid - /bin
* CentOS 7 - /bin
* Ubuntu 18 - /bin
* Ubuntu 16 - /bin

My guess is that /sbin is probably not anywhere other than CentOS6 and early fedora, though I don't have access to such systems to test on.

ChefAustin: Thanks for all the great feedback, suggestions. I'll take a whack at reverting the symlink addition and making the recommended changes.

ChefAustin: So I've gone back through the following distros/versions: 
- Fedora 24-32: All of these releases seem to have `/bin/udevadm` existing
- Ubuntu 16, 18. 20: All of these releases seem to have `/bin/udevadm` existing
- Debian 8, 9, 10: All of these releases seem to have `/bin/udevadm` existing
- CentOS 6, 7, 8: @jaymzh is correct insofar that CentOS 6 appears to be missing `/bin/udevadm` (and, instead, the bin exists at `/sbin/udevadm`). In CentOS 7 (and 8), this path exists. I was unable to get a working copy of the CentOS 5 ISO but this version reached EOL in March of 2017.

All that said, would y'all deem it acceptable to have the `value_for_platform` `default`'s key-value reflect `/bin/udevadm` and then a sole key for `'centos' => { '< 6.0' => '/sbin/udevadm'}` to cover this singular edge case?

Or would it be preferable to have each platform explicitly defined?

jaymzh: Yup. That's exactly what I think we should do. Thanks for the thorough checking!

ChefAustin: Great; thanks for confirming, @jaymzh!

I've pushed the changes and think I've covered all the discussed cases (but its still early and I'm not fully caffeinated yet, so double-check me on this commit 😆).

cookbooks/fb_systemd/recipes/udevd.rb
```diff
@@ -18,19 +18,26 @@
 # limitations under the License.
 #
 
+udevadm = value_for_platform(
+  'centos' => {
+    '< 6.0' => '/sbin/udevadm'
```
davide125, line 23: TIL operators are a thing. Looks like this goes back to at least Chef 12 (https://docs-archive.chef.io/release/12-0/dsl_recipe.html#id4) so it should be safe to use.

davide125, line 23: Add a trailing comma here for consistency

cookbooks/fb_systemd/recipes/udevd.rb
```diff
@@ -18,19 +18,26 @@
...
+  },
+  default => '/bin/udevadm',
```
davide125, line 25: `default` needs to be a string for this to work

jaymzh:
Both of those were caught by the travis linters/tests ... but since master is broken it wasn't obvious. :(

cookbooks/fb_systemd/recipes/udevd.rb
ChefAustin: > Add a trailing comma here for consistency

Fixed.

cookbooks/fb_systemd/recipes/udevd.rb
ChefAustin: > `default` needs to be a string for this to work

Fixed.

PR #126, opened 2020-05-09
Fix chef-cookbooks builds

jaymzh:
There's a handful of small issues. I want to fix them in a single
PR, otherwise you end up chasing down fixes that uncover other build
problems, so I'd like to see the build go green before wasting anyone's
time to import.

- [X] foodcritic timeout
    - Tell travis to wait up to 30 minutes
- [X] ubuntu18 - stunnel failing to start
    - Don't start the service with an empty config, stunnel won't take it
- [X] if `node['fb_base_settings']['_oscore_timers_rollout']`
    - guard this better.
- [X] Apache modules errors on Ubuntu
    - Remove configs we don't want, also create the symlink we expect
- [X] Apache mime errors
    - Create symlink
- [X] Apache start errors on CentOS7
    - Add unixd for "User", ssl and schmb for SSL stuff
    - Enable systemd for Type=notify
- [X] Rsyslog failures.
    - Two things were necessary - enabling Davide's undocumented fix, and increasing the retry delay significantly.... fortunately this only make runs longer when syslog fails
    - The "better" fix here, long-term, is probably to touch 
      /root/firstboot_os in our tests and enable this only firstboot, 
      when the race condition actually exists
- [X] gitignore
- [X] Documentation on debugging kitchen issues

It's not easy to stack a bunch of small diffs in github the way it is
in phabricator, or I would have, but I wanted to ensure the changes
would actually make all things pass. If you want to break them up into
internal PRs and close this that's fine, I don't need the credit, as
long as the tests are passing.

### DIFF
--- a/cookbooks/fb_apache/recipes/default.rb
+++ b/cookbooks/fb_apache/recipes/default.rb
@@ -120,6 +120,27 @@ end
   end
 end
 
+if node.debian? || node.ubuntu?
+  # CentOS makes this symlink to the right module dir, and we make assumptions
+  # it exists, so be sure to do the same on debian
+  link '/etc/apache2/modules' do
+    to '/usr/lib/apache2/modules'
+  end
+
+  # For reasons I don't understand on Ubuntu, Apache looks for mime.types in
+  # /etc/apache2/mime.types even though it's not configured to. So make a
+  # symlink
+  link '/etc/apache2/mime.types' do
+    to '/etc/mime.types'
+  end
+end
+
+# The package comes pre-installed with module configs, but we dropp off our own
+# in fb_modules.conf. Also, we don't want non-Chef controlled module configs.
+fb_apache_cleanup_modules 'doit' do
+  mod_dir moddir
+end
+
 template "#{moddir}/fb_modules.conf" do
   not_if { node.centos6? }
   owner 'root'
--- a/cookbooks/fb_apache/resources/cleanup_modules.rb
+++ b/cookbooks/fb_apache/resources/cleanup_modules.rb
@@ -1,0 +1,20 @@
+property :mod_dir, String
+
+action :manage do
+  allowed = [
+    "#{new_resource.mod_dir}/00-mpm.conf",
+    "#{new_resource.mod_dir}/fb_modules.conf",
+  ]
+  Dir.glob("#{new_resource.mod_dir}/*").each do |f|
+    next if allowed.include?(f)
+    if ::File.symlink?(f)
+      link f do
+        action :delete
+      end
+    else
+      file f do
+        action :delete
+      end
+    end
+  end
+end
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@
 *.rej
 Gemfile.lock
 .kitchen
+.bundle
+vendor
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,7 +31,7 @@ script:
   - shellcheck ./scripts/*
   - ./scripts/run_chefspec
   - ./scripts/run_rubocop .rubocop.yml
-  - ./scripts/run_foodcritic
+  - travis_wait 30 ./scripts/run_foodcritic
   - bundle exec kitchen verify $INSTANCE
 
 # Disabling deploy until we sort out Supermarket issues t39456893
--- a/README.kitchen_debug.md
+++ b/README.kitchen_debug.md
@@ -1,0 +1,8 @@
+# Debugging kitchen runs
+
+You can setup kitchen using the same commands as in `.travis.yml`, but once
+Chef runs you won't have access to connect, so modify
+`fb_sudo/attributes/default.rb` and uncomment the kitchen block.
+
+Then you can do `bundle exec kitchen login <INSTANCE>` after a failed
+run, and sudo will be passwordless so you can debug.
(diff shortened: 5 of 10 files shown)


jaymzh:
6 tests are passing now, more debugging to come :)

cookbooks/fb_apache/recipes/default.rb
```diff
@@ -106,6 +106,27 @@
...
+    to '/etc/mime.types'
+  end
+end
+
+# The package comes pre-installed with module configs, but we dropp off our own
+# in fb_modules.conf. Also, we don't want non-Chef controlled module configs.
+fb_apache_cleanup_modules 'doit' do
+  mod_dir moddir
+end
+
```
davide125, line 129: @jaininshah9 tested this on some fbit hosts and found we're dropping a lot of modules part of the Apache default config, which breaks stuf (e.g.: `mod_rewrite`). We could (and likely should) add those on our end, but I think we should at double check and update the default here to match the distro default.

jaymzh: I picked a default that lets the distro-default service start. Keeping up with whatever modules the distro disables/enables by default over time is probably a bit of a tall order, but I don't mind adding some...

Also keep in mind though that it's more intuitive for people to add modules than to remove them. And that additional modules can mean additional security vulns. So it seems to me less is more here?

davide125: @jaininshah9 can you share the list you have on your hosts, so we can make an informed decision? I tend to agree that we probably shouldn't add everything, but if we're missing something glaring we should fix it.

jaymzh: @davide125 - yeah that seems like a good idea. For rewrite in particular, I think it should not be in the default - it's something you have to turn-on per vhost anyway, so it seems like adding that module at the same time makes sense.

jaininshah9: Here is the list of enabled loaded modules (sorry for the long post, didn't know how else to post as paste wouldn't work here)

```
Loaded Modules:
 core_module (static)
 so_module (static)
 http_module (static)
 access_compat_module (shared)
 actions_module (shared)
 alias_module (shared)
 allowmethods_module (shared)
 auth_basic_module (shared)
 auth_digest_module (shared)
 authn_anon_module (shared)
 authn_core_module (shared)
 authn_dbd_module (shared)
 authn_dbm_module (shared)
 authn_file_module (shared)
 authn_socache_module (shared)
 authz_core_module (shared)
 authz_dbd_module (shared)
 authz_dbm_module (shared)
 authz_groupfile_module (shared)
 authz_host_module (shared)
 authz_owner_module (shared)
 authz_user_module (shared)
 autoindex_module (shared)
 cache_module (shared)
 cache_disk_module (shared)
 data_module (shared)
 dbd_module (shared)
 deflate_module (shared)
 dir_module (shared)
 dumpio_module (shared)
 echo_module (shared)
 env_module (shared)
 expires_module (shared)
 ext_filter_module (shared)
 filter_module (shared)
 headers_module (shared)
 include_module (shared)
 info_module (shared)
 log_config_module (shared)
 logio_module (shared)
 mime_magic_module (shared)
 mime_module (shared)
 negotiation_module (shared)
 remoteip_module (shared)
 reqtimeout_module (shared)
 rewrite_module (shared)
 setenvif_module (shared)
 slotmem_plain_module (shared)
 slotmem_shm_module (shared)
 socache_dbm_module (shared)
 socache_memcache_module (shared)
 socache_shmcb_module (shared)
 status_module (shared)
 substitute_module (shared)
 suexec_module (shared)
 unique_id_module (shared)
 unixd_module (shared)
 userdir_module (shared)
 version_module (shared)
 vhost_alias_module (shared)
 dav_module (shared)
 dav_fs_module (shared)
 dav_lock_module (shared)
 lua_module (shared)
 mpm_prefork_module (shared)
 proxy_module (shared)
 lbmethod_bybusyness_module (shared)
 lbmethod_byrequests_module (shared)
 lbmethod_bytraffic_module (shared)
 lbmethod_heartbeat_module (shared)
 proxy_ajp_module (shared)
 proxy_balancer_module (shared)
 proxy_connect_module (shared)
 proxy_express_module (shared)
 proxy_fcgi_module (shared)
 proxy_fdpass_module (shared)
 proxy_ftp_module (shared)
 proxy_http_module (shared)
 proxy_scgi_module (shared)
 proxy_wstunnel_module (shared)
 xml2enc_module (shared)
 proxy_html_module (shared)
 ssl_module (shared)
 systemd_module (shared)
 cgi_module (shared)
```

davide125: Thanks @jaininshah9. FYI you can use https://gist.github.com/ to make pastes for github.

cookbooks/fb_apache/resources/cleanup_modules.rb
```diff
@@ -0,0 +1,16 @@
...
+      link f do
+        action :delete
+      end
+    else
+      file f do
+        action :delete
+      end
+    end
+  end
+end
```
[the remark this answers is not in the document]
jaymzh: I'll fix this today. No need tod o the any, just need to compare them to `["{new_resource.mod_dir}/00-mpm.conf", "{new_resource.mod_dir}/fb_modules.conf"]`
or keep the array the same and do `.include?(File.basename(f))`. the former is more efficient. I'll update this later today.

PR #138, opened 2020-05-29
fb_systemd: don't unconditionally install systemd-journal-remote

jaymzh:
When possible, don't install the package unless enabled.


jaymzh:
@davide125 done.

PR #139, opened 2020-06-20
Move kitchen tests to Chef 13

jaymzh:
Because FB is on Chef 13 now. :)

### DIFF
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -1,8 +1,9 @@
 driver:
   name: docker
   privileged: true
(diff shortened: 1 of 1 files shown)


.kitchen.yml
```diff
@@ -2,7 +2,7 @@ driver:
   name: docker
   privileged: true
   # pin to chef 12 until our cookbooks are 13-safe
```
NaomiReeves, line 4: want to update this comment or remove it?

PR #144, opened 2020-06-27
fb_profile: handle bashrc properly on debian-like systems

jaymzh:
Per the very long discussion on the original submission of this,
this adds the agreed-upon handling of bashrc on debian-like systems to
add parity with rh-like systems.

It also moves the template back to the originally-submitted location
now that FB's tools should support this.


cookbooks/fb_profile/files/debian.bashrc
```diff
@@ -0,0 +1,107 @@
+# controlled by Chef, do not edit
```
davide125, line 1: I think this got munged, you should add the standard blurb like we have in the template

cookbooks/fb_profile/recipes/default.rb
```diff
@@ -27,12 +27,33 @@
...
+  # rubocop:disable ChefDeprecations/NodeMethodsInsteadofAttributes
   group node.root_group
+  # rubocop:enable ChefDeprecations/NodeMethodsInsteadofAttributes
+  # rubocop:enable Lint/UnneededCopDisableDirective
   mode '0644'
 end
+
+# Debian doesn't do the redhat make-sure-non-login-shells-get-aliases
+# So this is the bashrc from debian/ubuntu with that extra bit in there
+if node.debian? || node.ubuntu?
```
davide125, line 52: usually we prefer `only_if` for these

jaymzh: sort-of. `only_if` is generally preferred because it's runtime safe. In this case (1) the conditional is compile-time safe and (2) that file shouldn't *look* like it's being managed by Chef on RHEL-like machines, because that'll be confusing as hell to people.

jaymzh:
rebased, updated comment.

PR #146, opened 2020-06-27
fb_syslog: a variety of platform fixes

jaymzh:
* Make the default log file work for debian-like systmes
* Add in authlog as a default, it's in a default install anywhere I look
* Don't force default file permissions on everyone, use the API rather than hard-codi
ng it in the template
* Nuke rsyslog.d so people can't add configs outside of chef
* Fix some unittest

### DIFF
--- a/cookbooks/fb_syslog/recipes/default.rb
+++ b/cookbooks/fb_syslog/recipes/default.rb
@@ -74,6 +74,12 @@ template config_file do
   notifies :restart, "service[#{service_name}]"
 end
 
+directory '/etc/rsyslog.d' do
+  action :delete
+  recursive true
+  notifies :restart, "service[#{service_name}]"
+end
+
 service service_name do
   action :start
   subscribes :restart, 'package[rsyslog]'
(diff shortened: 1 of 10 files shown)


davide125:
Turns out we do have stuff in rsyslog.d in some places, so this will take a while to test internally.

jaymzh:
I'm not surprised. Took me a bit to roll out as well. But gotta cleanup sometime :)

ShahrozTanveer (approved):
* Great Work With Logs! 🚀👌

jaymzh:
Ping @davide125

davide125:
I'll try and get around it this week, but I'm oncall, so no promises :-)

jaymzh:
As it's February 12th, exactly 6 month ping on this one. I know a lots going on there, not expecting it today or anything, just wanna get it back on the radar.

cookbooks/fb_syslog/recipes/default.rb
```diff
@@ -74,6 +74,12 @@
   notifies :restart, "service[#{service_name}]"
 end
 
+directory '/etc/rsyslog.d' do
+  action :delete
+  recursive true
+  notifies :restart, "service[#{service_name}]"
+end
+
```
davide125, line 82: Let's do this in another PR please, and ideally put it behind an attribute gate.

davide125 (changes requested):

PR #147, opened 2020-06-27
fb_timers: some lint fixes

jaymzh:
Just small fix to keep cookstyle in its place


jaymzh:
this was not needed, bad sync.

PR #148, opened 2020-06-27
fb_tmpclean: lint fixes

jaymzh:
keep cookstyle in line


davide125:
Why not just turn off `ChefDeprecations/NodeMethodsInsteadofAttributes` until the bug is resolved upstream?

jaymzh:
That... *is* what I'm doing!

But then older versions of rubocop, or people not using Cookstyle yet (like facebook) will get errors about disabling a cop that doesn't exist. So I'm ALSO disabling that.

PR #152, opened 2020-09-21
Add support for reserved ranges to fb_users.

rb2k:
Summary:
There are cases where you don't manage a certain UID/GID, but you would like to make sure that nobody else accidentally tries to put their own user/group in that range either.

This code adds the ability to specify protected ranges which would fail validation if someone where to attempt putting a new UID/GID inside those ranges.

Differential Revision: D23817151

### DIFF
--- a/cookbooks/fb_users/libraries/default.rb
+++ b/cookbooks/fb_users/libraries/default.rb
@@ -32,6 +32,14 @@ module FB
           fail "fb_users[user]: User #{user} in UID map has a UID conflict"
         end
 
+        if defined?(RESERVED_UID_RANGES)
+          RESERVED_UID_RANGES.each do |identifier, range|
+            if range.include?(info['uid'])
+              fail "fb_users[user]: User #{user} in UID map is in the " +
+                "reserved range for '#{identifier}'"
+            end
+          end
+        end
         uids[info['uid']] = nil
       end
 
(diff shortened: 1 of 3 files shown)


cookbooks/fb_users/libraries/default.rb
```diff
@@ -40,7 +48,14 @@ def self._validate(node)
         if gids[info['gid']]
           fail "fb_users[group]: group #{group} in GID map has a GID conflict"
         end
-
+        if defined?(RESERVED_GID_RANGES)
```
jaymzh, line 51: you put a blank line between these above, so be consistent here. :)

cookbooks/fb_users/spec/default_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+              'action' => :add,
+            },
+          },
+        }
+      end
+
+      it 'should not fail if we did not specify a protected uid range' do
+        stub_const('FB::Users::UID_MAP', { 'testuser' => { 'uid' => 42 } })
+        stub_const('FB::Users::GID_MAP', { 'testgroup' => { 'gid' => 4242 } })
+        FB::Users._validate(node)
```
jaymzh, line 47: add an expectation here. `expect { FB::Users._validate(node) }.no raise_error(RuntimeError)`

jaymzh (changes requested):

cookbooks/fb_users/spec/default_spec.rb
rb2k: I would prefer to ensure it doesn't raise ANY error:

```
expect { FB::Users._validate(node) }.not_to raise_error
```

vs

```
expect { FB::Users._validate(node) }.not_to raise_error(RuntimeError)
```

Thoughts?

jaymzh: The default error is Runtime, so I don't think it matters much, but yeah I'm fine with that.

jaymzh (approved):
<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia3.giphy.com%2Fmedia%2F3oEjI5VtIhHvK37WYo%2Fgiphy.gif&token=nice" data-gitmeme-token="nice">nice</a>

PR #153, opened 2020-09-24
Add alternate acceptable platform value for def macos?

ChefAustin:
On some versions of Chef and macOS `node['platform'] returns `macos` instead of `mac_os_x`, this PR will allow either to be acceptable respones for `node.macos?` to return true.

### DIFF
--- a/cookbooks/fb_helpers/libraries/node_methods.rb
+++ b/cookbooks/fb_helpers/libraries/node_methods.rb
@@ -120,7 +120,7 @@ class Chef
     end
 
     def macos?
-      self['platform'] == 'mac_os_x'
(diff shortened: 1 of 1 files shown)


cookbooks/fb_helpers/libraries/node_methods.rb
```diff
@@ -120,7 +120,7 @@ def fedora_family?
     end
 
     def macos?
-      self['platform'] == 'mac_os_x'
+      self['platform'] == 'mac_os_x' || self['platform'] == 'macos'
```
davide125, line 123: prefer `%w{mac_os_x macos}.include?(self['platform'])`

ChefAustin: Good call. Fixed!

PR #157, opened 2020-11-13
fb_storage: when an array is mismatched don't include a missing filesystem on top

joshuamiller01:
Summary:
When an array is mismatched, as with incorrect members, or raid level, or a
partial stripe, report the required change as mismatches, and skip the rest
of array device evaluation, since missing filesystem comparisons are irrelavant
at that point, and because 'missing' has a different safety implication than
mismatched (e.g. data loss).

Differential Revision: D24960372

### DIFF
--- a/cookbooks/fb_storage/libraries/storage.rb
+++ b/cookbooks/fb_storage/libraries/storage.rb
@@ -807,21 +807,24 @@ module FB
           # and this will require us to nuke the array, so we'll need to
           # make a filesystem too
           mismatched_filesystems << device
+          next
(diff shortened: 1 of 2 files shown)


jaymzh:
Yeah this looks right to me. The array code needed more tests, thanks.

jaymzh (approved):

jaymzh:
@davide125 we need to get those tests working so we can tell if tests for this sort of stuff are passing. :/

joshuamiller01:
Merged internally.

PR #158, opened 2020-12-05
Include 'systemd-sysv' in default packages

jaymzh:
As @davide125 and I found out, if this package is out of date it can
cause weird thinks like networkd failing to start. Keep this up-to-date
with the rest.

This is a separate package in every distro I could find, so this should
work. Explicitly tested on debian, ubuntu 18.04, and centos7.


jaymzh:
chefstyle fixes added

jaymzh:
and markdownlint fixes

davide125 (approved):

PR #164, opened 2021-02-11
Add windows support to fb_ntp

jaymzh:
This PR adds basic windows support to fb_ntp.

The support is just for setting NTP servers and ensureing the service
is running. There are other things one can set and this could be
extended if desired, but it works well for now.

Note this uses powershell_exec which I think requires Chef 16. It could
be moved to powershell_out if necessary, though that is much, *much*
slower.

This also has a chefstyle autocorrect which does implicit rescue which
requires Chef 13, but I _think_ the last of 12 is dead at FB so that
should be fine, but someone should doublecheck.

### DIFF
--- a/cookbooks/fb_ntp/recipes/default.rb
+++ b/cookbooks/fb_ntp/recipes/default.rb
@@ -30,8 +30,6 @@ service_name = value_for_platform(
 
 if node.macos?
   include_recipe 'fb_ntp::macosx'
-elsif node.windows?
-  fb_ntp_windows_config 'doit'
 end
 
 whyrun_safe_ruby_block 'enforce ACL hardening' do
@@ -88,16 +86,25 @@ end
 fb_systemd_override 'local' do
   only_if { node.systemd? }
   unit_name 'ntpd.service'
-  content({
-            'Service' => {
-              'Restart' => 'always',
-            },
-          })
+  content(
+    {
+      'Service' => {
+        'Restart' => 'always',
+      },
+    },
+  )
 end
 
 service service_name do
   not_if { node.macos11? }
   action [:enable, :start]
+end
+
+# You can't configure the service unless it's up, so this has to be
+# *after* the `service` call above. It does need to notify the service
+# as the commands make the change live and persistent.
+if node.windows?
+  fb_ntp_windows_config 'doit'
 end
 
 # ntpdate is a service that should only run at boot, as that is the only time
--- a/cookbooks/fb_ntp/recipes/default.rb
+++ b/cookbooks/fb_ntp/recipes/default.rb
@@ -44,16 +44,23 @@ whyrun_safe_ruby_block 'enforce ACL hardening' do
     ]
 
     # Resolve chosen timesources and allow them
+    #
+    # implicit-begin is a function of ruby2.5 and later, but we still
+    # support 2.4, so.... until then
+    # rubocop:disable Style/RedundantBegin
     node['fb_ntp']['servers'].each do |host|
-      ips = Resolv.getaddresses(host)
-      ips.each do |ip|
-        dash6 = ''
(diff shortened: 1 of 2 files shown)


cookbooks/fb_ntp/resources/windows_config.rb
```diff
@@ -0,0 +1,39 @@
...
+  want = node['fb_ntp']['servers']
+  have = config['TimeProviders']['NtpServer'].split(',')
+  if Set.new(want) != Set.new(have)
+    Chef::Log.info(
+      'fb_ntp_windows_config: Changing NTP servers from ' +
+      "#{have.join(', ')} to #{want.join(', ')}",
+    )
+    set_ntp_servers
+  else
+    Chef::Log.debug('fb_ntp_windows_config: NTP servers are correct')
```
davide125, line 37: nit: the convention for logging in LWRPs is to use `fb_ntp[windows_config]`

cookbooks/fb_ntp/resources/windows_config.rb
```diff
@@ -0,0 +1,39 @@
+action_class do
+  def get_current_config
+    config = {}
+    section = nil
+    s = powershell_exec('w32tm /query /configuration')
```
davide125, line 5: can/should this be moved to a ohai plugin?

davide125:
You'll also need to update `metadata.rb`

cookbooks/fb_ntp/resources/windows_config.rb
jaymzh: No, resources should always check the state at runtime.

cookbooks/fb_ntp/resources/windows_config.rb
jaymzh: yes, good call.

cookbooks/fb_ntp/recipes/default.rb
```diff
@@ -43,16 +47,14 @@
...
-      rescue Resolv::ResolvError
-        Chef::Log.warn("fb_ntp: failed to resolve #{host}, skipping")
+      ips = Resolv.getaddresses(host)
+      ips.each do |ip|
+        dash6 = ''
+        dash6 = '-6 ' if IPAddr.new(ip).ipv6?
+        acl_entries << "restrict #{dash6}#{ip} ##{host}"
       end
+    rescue Resolv::ResolvError
+      Chef::Log.warn("fb_ntp: failed to resolve #{host}, skipping")
```
malmond77, line 57: This changes the error handling to skip single servers that don't resolve, to skipping all servers past the first one that fails. Why the change in semantics here?

jaymzh: rubocop did this automatically. :( seems like a bug, you're right. :/

jaymzh: Nope, wait, you're misunderstanding the code. The implicit rescue there doesn't stop the loop... it rescues. The code is identical:

```
[phil@rider ~]$ cat /tmp/test.rb 
#!/usr/bin/ruby

(0..9).each do |i|
  puts i
  raise if i == 5
rescue
  puts 'i am at 5'
end
[phil@rider ~]$ /tmp/test.rb
0
1
2
3
4
5
i am at 5
6
7
8
9
```

That code is correct.

jaymzh:
> You'll also need to update `metadata.rb`

for what? this doesn't have a supports.

jaymzh:
should be good now @davide125

cookbooks/fb_ntp/recipes/default.rb
malmond77: This sort of thing seems to be par for the course for Ruby. Alright, TIL then.

davide125 (approved):

PR #165, opened 2021-02-11
fb_helpers fixups

jaymzh:
* The unittest assumed the machine it was running on was PST, oops.
* The README didn't pass MDL rules
* Some other chefstyle fixes

### DIFF
--- a/cookbooks/fb_helpers/libraries/node_methods.rb
+++ b/cookbooks/fb_helpers/libraries/node_methods.rb
@@ -803,11 +803,17 @@ class Chef
       return default if path.nil?
 
       node_path = path.split(delim)
+      # implicit-begin is a function of ruby2.5 and later, but we still
+      # support 2.4, so.... until then
+      # rubocop:disable Style/RedundantBegin
       node_path.inject(self) do |location, key|
-        location.attribute?(key.to_s) ? location[key] : default
-      rescue NoMethodError
-        default
+        begin
+          location.attribute?(key.to_s) ? location[key] : default
+        rescue NoMethodError
+          default
+        end
       end
+      # rubocop:enable Style/RedundantBegin
     end
   end
 end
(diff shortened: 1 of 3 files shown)


cookbooks/fb_helpers/libraries/node_methods.rb
```diff
@@ -689,7 +689,7 @@ def root_user
 
     def root_group
       value_for_platform(
-        %w{openbsd freebsd mac_os_x macos} => { 'default' => 'wheel' },
```
davide125, line 692: @rb2k can you check that this is fine on bigsur too?

rb2k: It depends on which version of chef you run.
Newer versions will map macOS to always be mac_os_x:

https://github.com/chef/ohai/blob/f26c3ce26c63eb06d714073f920b1f75c4feef25/lib/ohai/plugins/darwin/platform.rb#L35

older versions  (e.g. chef 14) will not:
https://github.com/chef/ohai/blob/14-stable/lib/ohai/plugins/darwin/platform.rb#L30

This was the PR: https://github.com/chef/ohai/pull/1470

jaymzh:
should be good now @davide125

malmond77:
The code looks good, but the tests show a couple of different errors: Ruby 2.4 shows a syntax error, and 2.5/2.6 show:

```
rspec ./cookbooks/fb_network_scripts/spec/gated_spec.rb:34 # fb_network_scripts::spec centos7 (7.3.1611) should try to update the template when nw changes are allowed
```

The latter seems related to work done by @joshuamiller01 and repros on master. That shouldn't block this change, but the syntax error has me concerned.

jaymzh:
Good catch @malmond77 . Can someone over there confirm for me what version of ruby your version of chef is on? If it's still 2.4, I'll make sure this is compatible, if not we should drop the tests

malmond77:
Alas 2.4 is still a thing. I dug around Github for a moment to see if the issue is a regression or not in this code. I think it is because while https://github.com/facebook/chef-cookbooks/actions/runs/558973593 also failed, it didn't fail in the same way.

jaymzh:
K @malmond77 then I'll fix it here and in fb_ntp where we did the same thing.

PR #167, opened 2021-03-22
allow setting `homedir_mode` from `fb_users` api

NaomiReeves:
Summary:
The `user` resource doesn't automatically allow for the management of mode of the user's home directory. On CentOS hosts the homedir mode is usually managed by the defaults set in `/etc/login.defs`, `UMASK` and `HOME_MODE`, but api users may need to configure this for individual users. By setting `'manage_home' => true`, `fb_users` will now wholly own the user's homedir, and it is possible to configure the permissions on that home directory by specifying something like:
```
node.default['fb_users']['users']['jane'] = {
  'shell' => '/bin/zsh',
  'gid' => 'users',
  'manage_home' => true,
  'home' => '/var/foo/jane',
  'homedir_mode' => '0755',
  'action' => :add,
}
```

Open question: if the user has `'action' => :delete`, should the directory be cleaned up/removed? Or should that only be done if  `'action' => :delete` && `'manage_home' => true`? It could be possible to use the data from `node['passwd']` to identify the `dir` for the user and clean that target up.

Differential Revision: D27247064

### DIFF
--- a/cookbooks/fb_users/README.md
+++ b/cookbooks/fb_users/README.md
@@ -9,9 +9,9 @@ Requirements
 Attributes
 ----------
 * node['fb_users']['user_defaults']
+* node['fb_users']['user_defaults']['gid']
 * node['fb_users']['user_defaults']['manage_home']
 * node['fb_users']['user_defaults']['shell']
-* node['fb_users']['user_defaults']['gid']
 * node['fb_users']['users']
 * node['fb_users']['users'][$USER]
 * node['fb_users']['groups']
@@ -20,6 +20,7 @@ Attributes
 Usage
 -----
 ### Consistent data vs dynamic data
+
 This cookbook draws a hard distinction between information that should not
 be changable across an organization and information that should.
 
@@ -29,6 +30,7 @@ per-system like if the user is on the system and what groups they are a part
 of.
 
 ### Pre-req: Initializing Consistent data with UID_MAP AND GID_MAP
+
 UIDs, GIDs, and (optionally) user/group system flags or "comments" are
 considered consistent data by `fb_users`. These exist in a single map. They do
 *not* effect what users or groups are installed on a system and are just a
@@ -101,7 +103,31 @@ group, and provide a `comment`.
 
 All other values must be set in the node object.
 
+### Reserved UID / GID ranges
+
+If you want to protect certain UID / GID ranges from being used across your
+managed system, you can add them to `RESERVED_UID_RANGES` and `RESERVED_GID_RANGES`.
+This can be useful if 3rd party software creates those users/groups.
+The key in the `Hash` will be a printable identifier, the value will be an object
+that responds to `.include?()`, so usually `Range` or `Array`.
+
+```
+module FB
+  class Users
+    RESERVED_UID_RANGES = {
+      'systemd dynamic users' => 61184..65519,
+      'project foo service users' => [52231, 52233],
+    }
+    RESERVED_GID_RANGES = {
+      'project x service groups' => 30100..30200
+      'acme corp service groups' => [30100, 30442]
+    }
+  end
+end
+```
+
 ### Users and Groups
+
 Users and groups are setup using the `users` and `groups` hashes and are
 straight forward:
 
@@ -124,6 +150,8 @@ before group membership is managed.
 By design, we do not accept all values. Here are the values we do accept:
 * `gid`
 * `home`
+* `homedir_group`
+* `homedir_mode`
 * `manage_home`
 * `password`
 * `shell`
@@ -138,27 +166,6 @@ and then if later recipes add them, they'll be added, otherwise they'll be
 automatically cleaned up.
 
 Also see `initialize_group` helper below.
-
-If you want to protect certain UID / GID ranges from being used across your
-managed system, you can add them to `RESERVED_UID_RANGES` and `RESERVED_GID_RANGES`.
-This can be useful if 3rd party software creates those users/groups.
-The key in the `Hash` will be a printable identifier, the value will be an object
-that responds to `.include?()`, so usually `Range` or `Array`.
-
-```
-module FB
-  class Users
-    RESERVED_UID_RANGES = {
-      'systemd dynamic users' => 61184..65519,
-      'project foo service users' => [52231, 52233],
-    }
-    RESERVED_GID_RANGES = {
-      'project x service groups' => 30100..30200
-      'acme corp service groups' => [30100,30442]
-    }
-  end
-end
-```
 
 ### Passwords in data_bags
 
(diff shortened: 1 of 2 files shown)


cookbooks/fb_users/resources/default.rb
```diff
@@ -120,6 +122,15 @@
       password pass if pass
       action :create
     end
+
+    if manage_homedir
+      directory homedir do
+        owner mapinfo['uid']
+        group ::FB::Users::GID_MAP[pgroup]['gid'].to_i
+        mode homedir_mode if homedir_mode
```
jaymzh, line 130: Not everyone's homedir will necessary be their primary group. For example my primary group could be `os` but the company could want all homedirs to be group-owned by `pe`, or vise-versa.

Since `manage_homedir` defaults to false, this probably OK. But something to think about. I would consider maybe a `homedir_group` config, then fall back to user's GID if not set.

cookbooks/fb_users/README.md
```diff
@@ -9,9 +9,10 @@ Requirements
 Attributes
 ----------
 * node['fb_users']['user_defaults']
+* node['fb_users']['user_defaults']['gid']
+* node['fb_users']['user_defaults']['homedir_mode']
```
jaymzh, line 13: I don't think this makes sense in `user_defaults`. If you want to change the _default_, then you should change `login.defs`. If you want to change for a _specific_user_, then you'd want it in this API under the specific `users` entry (as you have below).

jaymzh:
> Open question: if the user has 'action' => :delete, should the directory be cleaned up/removed? Or should that only be done if 'action' => :delete && 'manage_home' => true? It could be possible to use the data from node['passwd'] to identify the dir for the user and clean that target up.

IMO if `manage_home` is `true`, then action `:delete` should nuke it. If people want different behavior for different actions they can always do that in code, but `manage_home` should mean the home is managed according to the action specified....

with a few exceptions:
* Users often get a homedirectory of `/` or `/tmp`. I have made the mistake of auto-nuking those before, I don't recommend it. :)
* For sanity, probably not worth deleting homedirs of any "system" users... but I could be convinced otherwise
* Perhaps a compromise (though I don't know if I actually like this or not), is we only delete the homedir if the homedir is actually specified in the node attr.

cookbooks/fb_users/README.md
davide125: This brings a good point -- should we have an API in `fb_users` to manage `login.defs` ?

cookbooks/fb_users/resources/default.rb
davide125: I like the `homedir_group` idea here

cookbooks/fb_users/README.md
jaymzh: I could argue it goes somewhere else, but I could also argue it goes here. If we wanna put it here I certainly have no objections.

NaomiReeves: I'm going to address managing `/etc/login.defs` in a separate change

NaomiReeves:
> IMO if `manage_home` is `true`, then action `:delete` should nuke it. If people want different behavior for different actions they can always do that in code, but `manage_home` should mean the home is managed according to the action specified....

Chef's `user` resource will try to do the right thing and clean up home dirs if a `user` resource has `manage_home` set to `true` and `action :remove`. I'm going to just use that logic and allow the resource to do the right thing. If someone needs more complicated logic for a particular user, they can define that separately in code and set `manage_home` to `false` so the user is still cleaned up but the home dir is left alone.

jaymzh:
> Chef's user resource will try to do the right thing and clean up home dirs if a user resource has manage_home set to true and action :remove. I'm going to just use that logic and allow the resource to do the right thing. If someone needs more complicated logic for a particular user, they can define that separately in code and set manage_home to false so the user is still cleaned up but the home dir is left alone.

Yeah I think that's the right call. `useradd` has lots of logic to get this right, and Chef leverages that, and we should leverage Chef leveraging that.

cookbooks/fb_users/resources/default.rb
```diff
@@ -120,6 +122,15 @@
       password pass if pass
       action :create
     end
+
+    if manage_homedir
+      directory homedir do
+        owner mapinfo['uid']
+        group ::FB::Users::GID_MAP[pgroup]['gid'].to_i
```
jaymzh, line 129: as mentioned before, this should probably be `info['homedir_group'] || ::FB::Users::GID_MAP[pgroup]['gid'].to_i`

jaymzh (changes requested):

cookbooks/fb_users/README.md
```diff
@@ -101,6 +101,28 @@ group, and provide a `comment`.
...
+```
+module FB
+  class Users
+    RESERVED_UID_RANGES = {
+      'systemd dynamic users' => 61184..65519,
+      'project foo service users' => [52231, 52233],
+    }
+    RESERVED_GID_RANGES = {
+      'project x service groups' => 30100..30200
+      'acme corp service groups' => [30100,30442]
```
jaymzh, line 120:
```suggestion
      'acme corp service groups' => [30100, 30442]
```

cookbooks/fb_users/README.md
```diff
@@ -200,21 +206,23 @@ in `GID_MAP`.
...
+will be automatically cleaned up during the chef run. To automatically clean up
+a user's home directory while removing the user from the system, leave
+`manage_home` set to `true`.
 
+### Helper methods
 This cookbook provides a few helper methods for your convenience:
 
 #### FB::Users.initialize_group(node, groupname)
-
 Since initializing a group nicely involves setting `members` to an empty array
```
jaymzh, line 219: Please don't delete these empty lines. First, MDL says there should be a blank line. Second, having it allows one to `gqap` in vim to autoformat/wrap. Otherwise it gets wrapped *with* the title/heading.

cookbooks/fb_users/resources/default.rb
```diff
@@ -69,15 +68,11 @@
...
-      user username do # ~FB014
-        action :remove
-      end
-      next
-    end
+    # helper variables
     mapinfo = ::FB::Users::UID_MAP[username]
     pgroup = info['gid'] || node['fb_users']['user_defaults']['gid']
     homedir = info['home'] || "/home/#{username}"
+    homedir_group = info['homedir_group'] || pgroup
```
jaymzh, line 75: I believe pgroup is a gid, but homedir_group is a group, you'll need to translate that one.

cookbooks/fb_users/resources/default.rb
```diff
@@ -120,6 +125,15 @@
       password pass if pass
       action :create
     end
+
+    if manage_homedir
+      directory homedir do
+        owner mapinfo['uid']
+        group ::FB::Users::GID_MAP[homedir_group]['gid'].to_i
```
jaymzh, line 132: homedir_group if I'm following, may already be a GID, per above.

jaymzh (commented):
a few comments

cookbooks/fb_users/resources/default.rb
NaomiReeves: I don't think `pgroup` is a gid at this point. It's translated later in the user resource at line 119

jaymzh (approved):
LGTM.

<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia3.giphy.com%2Fmedia%2Fta83CqOoRwfwQ%2Fgiphy.gif&token=shipit" data-gitmeme-token="shipit">shipit</a>="https://gitme.me/image?url=https%3A%2F%2Fmedia3.giphy.com%2Fmedia%2Fta83CqOoRwfwQ%2Fgiphy.gif&token=shipit" data-gitmeme-token="shipit"></a>

Issue #79, opened 2019-10-04
fb_zfs fails on every single supported distro


jaymzh:
all centos and all debian/ubuntus fail for something equivalent to:
```
No candidate version available for spl, spl-dkms, zfs-dkms, zfsutils-linux
```

nicolasgoudard:
UP

jaymzh:
@nicolasgoudard after talking to some folks it sounds like this cookbook is probably just gonna get deleted. No one is using it anymore.

@davide125 is that still the plan?

davide125:
Yes, this has been broken and unsupported for years, and I don't have any way to test it nowadays. Unless someone that's actually using ZFS wants to fix it and keep it working, it'll get deleted soon.

nicolasgoudard:
Hi, thx. It's ok now, I follow this tutorial https://openzfs.github.io/openzfs-docs/Getting%20Started/Debian/Debian%20Buster%20Root%20on%20ZFS.html

`sudo vi /etc/apt/sources.list`

Then add 

```
deb http://deb.debian.org/debian buster main contrib
deb http://deb.debian.org/debian buster-backports main contrib
```

PR #168, opened 2021-04-01
fb_zfs: delete deprecated cookbook

davide125:
Summary: `fb_zfs` has been effectively unmaintained for several years, and has long standing issues like https://github.com/facebook/chef-cookbooks/issues/79 that aren't trivial to fix without a real deployment to test it on.

Reviewed By: nikitakoshikov

Differential Revision: D23934240


jaymzh (changes requested):
please remove the references in test_services cookbook

PR #169, opened 2021-06-03
fb_users: windows fixes, removing user fixes

jaymzh:
This primarily fixes Windows support by doing two things:

1. Adding a new `set_passwords_on_windows` attribute which people will
almost certainly want to set to `false`, but I've defaulted to `true`
for backwards compatibility. The only passwords that you can pass to
Chef for Windows users are plain-text ones which not only is silly, but
also would be different than that user on other platforms. So if you
have a unix hash, it'll set the password to garbage on windows, so you
just want to skip password-setting on windows. See README.md for extra
details.

2. Skips a few things on Windows that don't make sense like checking for
NFS homedirs, or primary-group bootstrapping

In addition this is a slightly more comprehensive fix to removing users
that aren't in the UID map. [This
commit](https://github.com/facebook/chef-cookbooks/commit/d38e9a8a8be42929ee5be93b10dfc9dee5a683f5#diff-bd1cf9dcc4317e6497de436c3237684effa5d239e42d94d20bcfb67544e564f4)
worked-around it, but we should never have been passing `uid` to the
resource when we removed a user anyway, so rather than conditionally
pass it, we just don't pass it. This is how I've had it locally for
quite a while and it works well. Further, that change accidentally
allowed 'uid' to be passed in the node for :remove, which while it would
be harmless as we never look at it, we still shouldn't allow.


jaymzh:
rebased but NOT tested... will try to do that.

jaymzh:
OK this is now tested.

cookbooks/fb_users/resources/default.rb
```diff
@@ -136,9 +149,12 @@
...
       comment mapinfo['comment'] if mapinfo['comment']
-      password pass if pass
-      if FB::Version.new(Chef::VERSION) >= FB::Version.new('15')
-        secure_token info['secure_token'] unless info['secure_token'].nil?
+      if pass && set_passwords
+        password pass
+      end
+      if FB::Version.new(Chef::VERSION) >= FB::Version.new('15') &&
+          !mapinfo['secure_token'].nil?
+        secure_token info['secure_token']
```
joshuamiller01, line 157: mapinfo['secure_token'], right?

joshuamiller01: Hmmm ... we checked if `info` was nil before, now checking `mapinfo`; not sure what's intended here.

jaymzh: yeah this is weird, lemme fix

jaymzh: Fixed. Nice catch.

PR #170, opened 2021-06-03
Fix CentOS6 kitchen tests

jaymzh:
The URL for the epel for 6 moved. We probably want to drop 6 tests,
but at least for now this gets kitchen running.


davide125 (approved):

jaymzh:
@davide125 - you'll have to import and land, those files are all shipit-ified, IIRC.

PR #171, opened 2021-06-18
Add `notifies` to fb_users

jaymzh:
This adds `notifies` to the fb_users Attribute API, which should
eliminate the need for short-term fix of moving resources up into the
root run_context which has broken other things. And then when _that_
is cone we can remove the work around for that fix to notify Ohai.

I will follow up with a second PR to remove the `with_run_context`
calls, but this will have to get merged, then all users add the notifs
they want, then we can remove the calls.

This also pulls in the fix to unit tests from #169 which I had to pull
in to even be able to _run_ tests. :/

Also, test tests never worked, gruops don't have comments:

```
fb_users::default centos7 (7.3.1611) with user_defaults manage group creates the group with the values specified in GID_MAP
fb_users::default mac_os_x (10.12) with user_defaults manage group creates the group with the values specified in GID_MAP
```

So I nuked them.

### DIFF
--- a/cookbooks/fb_users/README.md
+++ b/cookbooks/fb_users/README.md
@@ -156,6 +156,7 @@ By design, we do not accept all values. Here are the values we do accept:
 * `password`
 * `shell`
 * `secure_token`
+* `notifies`
 
 They are all optional, see the next section for how default values for these
 work.
@@ -169,6 +170,33 @@ and then if later recipes add them, they'll be added, otherwise they'll be
 automatically cleaned up.
 
 Also see `initialize_group` helper below.
+
+The `notifies` is a hash with of notifies, where each notify is a hash with
+`resource`, `action`, and `timing`. It looks like this:
+
+```ruby
+node.default['fb_users']['users']['zerocool'] = {
+  ...
+  'notifies' => {
+    'restart foo if uid changes' => {
+      'resource' => 'service[foo]',
+      'action' => 'restart',
+      'timing' => 'delayed',
+    }
+  }
+}
+```
+
+The name of the notifier is inconsequential, and is only there to make it
+easy to reach in and modify a notification later in your run list. The action
+can be a string or a symbol, we will do the right thing.
+
+When such a notification is setup, we will run it regardless of if the user or
+group is being added, modified, or removed. The only time we don't notify is
+if the homedir is being changed.
+
+Note that to enable subscribes, these sub-resources were added to the root
+run context, but will be moved back to the subresource run context shortly.
 
 ### Passwords in data_bags
 
@@ -221,12 +249,6 @@ To remove a user or group, set the `action` to `:delete` and the user or group
 will be automatically cleaned up during the chef run. To automatically clean up
 a user's home directory while removing the user from the system, leave
 `manage_home` set to `true`.
-
-### Notifications for users or groups
-
-The `user` and `group` resources used within this cookbook's custom resource
-will run at the root `run_context` in order to allow other resources in the
-chef run to *subscribe* to a specific user or group being updated.
 
 ### Helper methods
 
--- a/cookbooks/fb_users/resources/default.rb
+++ b/cookbooks/fb_users/resources/default.rb
@@ -40,16 +40,22 @@ action :manage do
       next
     end
 
+    info = node['fb_users']['groups'][grp]
+
     # We may not have this group if it's a remote one, so check we do and
     # that it's set to create
-    if node['fb_users']['groups'][grp] &&
-        node['fb_users']['groups'][grp]['action'] &&
-        node['fb_users']['groups'][grp]['action'] != :delete
-
+    if info && info['action'] && info['action'] != :delete
       group "bootstrap #{grp}" do # ~FB015
         group_name grp
         gid ::FB::Users::GID_MAP[grp]['gid']
         action :create
+        # we'll likely modify the group below, but if it has no members and no
+        # comment, then we won't, so lets hook up the notifies in both places
+        # just in case
+        info['notifies']&.each_value do |notif|
+          timing = notif['timing'] || 'delayed'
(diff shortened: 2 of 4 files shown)


cookbooks/fb_users/README.md
```diff
@@ -170,6 +171,27 @@ automatically cleaned up.
 
 Also see `initialize_group` helper below.
 
+The `notifies` is a hash that looks like this:
+
+```ruby
+node.default['fb_users']['users']['zerocool'] = {
+  'restart foo if uid changes' => {
```
NaomiReeves, line 178: shouldn't this be `notifies` based on `info['notifies']` on line 55 in the resource?
```suggestion
  'notifies' => {
    'restart foo if uid changes' => {
```

cookbooks/fb_users/README.md
```diff
@@ -170,6 +171,27 @@ automatically cleaned up.
...
+    'resource' => 'service[foo]',
+    'action' => 'restart',
+  }
```
NaomiReeves, line 181: this should probably have a timer option too, specifically since the default timer is `:delayed` and most of the use cases I can think of require `:immediately`
```suggestion
    'timer' => :immediately,
  }
```

cookbooks/fb_users/resources/default.rb
```diff
@@ -40,16 +40,21 @@
...
+    if info && info['action'] && info['action'] != :delete
       group "bootstrap #{grp}" do # ~FB015
         group_name grp
         gid ::FB::Users::GID_MAP[grp]['gid']
         action :create
+        # we'll likely modify the group below, but if it has no members and no
+        # comment, then we won't, so lets hook up the notifies in both places
+        # just in case
+        info['notifies']&.each_value do |notif|
+          notifies notif['action'], notif['resource']
```
NaomiReeves, line 56:
```suggestion
          notifies notif['action'], notif['resource'], notif['timer']
```

cookbooks/fb_users/recipes/default.rb
```diff
@@ -19,6 +19,12 @@
 # limitations under the License.
 #
 
+if defined?(ChefSpec)
+  file 'test resource' do
```
NaomiReeves, line 23: can we just define this in the spec vs directly in the recipe?

NaomiReeves (changes requested):

cookbooks/fb_users/recipes/default.rb
jaymzh: I can find no way to do that - if yo know of one, I'm all ears. The `chef_run do ... end` doesn't have the recipe DSL in it, not does the `converge do ... end`. Most examples I find on the internet either do what I"m doing here, or they make a test recipes, that the include_recipe's the real recipe and they run the specs on the test recipe, but that seems worse to me.

cookbooks/fb_users/README.md
jaymzh: I added it, but I called it 'timing' to not be confused wiuth systemd timers.

cookbooks/fb_users/recipes/default.rb
NaomiReeves: I was thinking about something like this: https://github.com/chefspec/chefspec/blob/master/examples/recipe_block/spec/recipe_block_spec.rb#L16-L29

cookbooks/fb_users/resources/default.rb
```diff
@@ -40,16 +40,22 @@
...
+    if info && info['action'] && info['action'] != :delete
       group "bootstrap #{grp}" do # ~FB015
         group_name grp
         gid ::FB::Users::GID_MAP[grp]['gid']
         action :create
+        # we'll likely modify the group below, but if it has no members and no
+        # comment, then we won't, so lets hook up the notifies in both places
+        # just in case
+        info['notifies']&.each_value do |notif|
+          timing = notif['timing'] || 'delayed'
```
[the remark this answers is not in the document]
jaymzh: Sure we will! Chef will try to run the `user foo` resource and it'll blow up at compile time, and Chef will show both the source code as well as the compiled resource of `user foo do ...`, allowing the user to see exactly what's wrong. I think the normal fail here is fine.

cookbooks/fb_users/recipes/default.rb
jaymzh: I see how you got there, but I don't think that works quite like you think it does,

We already _have_ a `recipe` block in the test, at the very top:
https://github.com/facebook/chef-cookbooks/blob/master/cookbooks/fb_users/spec/resource_spec.rb#L24

The `recipe` block can either be an inline recipe, like the one you link to, or it can be a recipe from your cookbook you want to test, as it is in this case:

```
recipe 'fb_users::default' do |tc|
```

But more importantly, we actually override the `recipe` keyword in the fbspec stubs so that it sets up all the tons and tons of boilerplate you have to setup to get all your platforms and recipes setup in chefspec which is really cumbersome, so `recipe` for us means something a bit different (roughly: loop over all our platforms - module `include` and `exclude` - and then setup the proper recipe block for each fauxhai profile).

joshuamiller01: You can stitch in a 'real' recipe (e.g. fb_users::spec) at the `converge` statement, so have something like: `converge('fb_users::spec', described_recipe)`.  Having the separate "unit test" recipe feels lot cleaner to me versus polluting the primary implementation recipe.

jaymzh: Oh neat, thanks @joshuamiller01 ! I'll give that a shot.

jaymzh:
Ugh. Some test started failing in the interim

jaymzh:
I finally figured out those two failing tests... groups don't have comments. Not a thing...

Unclear why they pass on CI. Will drop that.

jaymzh:
There we go. Fixed the test that are on master for ya too :)

PR #173, opened 2021-06-18
fb_consul: Allow pinning of Consul version

marc-sensenich:
Summary: Add the `version` attribute to the fb_consul API to allow for
consumers to pin to a specified Consul version.

Example output from a Chef run with
`node.default['fb_consul']['version']` not set

```sh
...
Recipe: fb_consul::default
  * apt_package[upgrade consul] action upgrade (up to date)
  * apt_package[pin consul] action install (skipped due to not_if)
...
```

Example output from a Chef run with
`node.default['fb_consul']['version']` set

```sh
...
Recipe: fb_consul::default
  * apt_package[upgrade consul] action upgrade (skipped due to only_if)
  * apt_package[pin consul] action install (up to date)
...
```

### DIFF
--- a/cookbooks/fb_consul/README.md
+++ b/cookbooks/fb_consul/README.md
@@ -8,6 +8,8 @@ Requirements
 Attributes
 ----------
 * node['fb_consul']['enable']
+* node['fb_consul']['manage_packages']
+* node['fb_consul']['version']
 * node['fb_consul']['config']
 * node['fb_consul']['services']
 * node['fb_consul']['checks']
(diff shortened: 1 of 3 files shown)


NaomiReeves:
Hey @marc-sensenich sorry for the delay in responding to this. Pinning package versions isn't a pattern our cookbooks support, but we do allow cookbook users to set:
```
node.default['fb_consul']['manage_packages'] = false
```
and take over managing the package in their wrapper cookbook.

If that pattern doesn't work for you, could you give a bit more context about your workflow or use case to help justify a deviation to our cookbook api patterns?

jaymzh:
We've dropped `version` internally now.

PR #174, opened 2021-07-22
Add macos12 example to readme

homebysix:
The `macos12` function is already in node_methods.rb, but not yet in the readme.


jaymzh (approved):

jaymzh:
(The kitchen tests will fail, they're known to be broken.)

PR #177, opened 2021-09-07
fb_network_scripts: start making gated_template tests pass

jaymzh:
They don't pass yet, but this gets them a lot closer. I mostly
made this for @davide125 to take over and finish.


davide125:
cc @anitazha who is refactoring this logic for networkd

anitazha:
Yea I'm working on moving this to fb_helpers so I can reuse it for fb_networkd. Do you mean just the github tests? It does appear to pass if I do use fb_chefspec on the commandline.

jaymzh:
In this case the tests don't pass on GH because - at least in part - they run in a container which doesn't pull in the systemd stuff which then lets all the rest of this crap break.

scottruss (changes requested):
Because of other changes that have happened after a rebase this diff only adds the modprobe to the metadata.

jaymzh:
@scottruss still necessary to fix the tests. Are you asking for a rebase, or are you saying you're gonna add that and I should drop this?

jaymzh:
@scottruss rebased.

cookbooks/fb_helpers/recipes/spec.rb
```diff
@@ -20,6 +20,11 @@
 
 # This recipe is only for running ChefSpec tests
 if defined?(ChefSpec)
+  # network scripts uses the fb_modprobe_module resource which depends
+  # on the execute[load modules] resource in systemd
+  include_recipe 'fb_systemd'
+  include_recipe 'fb_modprobe'
+  include_recipe 'fb_network_scripts'
```
joshuamiller01, line 27: Pulling in a bunch of largely unrelated recipes (and their dependencies) will make this horribly brittle, and make the deps explode for fb_helpers.

Backing up - what's actually breaking here?  I can't infer what systemd has to do with this spec.  And I'm failing to find the original error which is showing up in GH

jaymzh: Well the fb_systemd is for:
```
       expected Chef::Exceptions::UserIDNotFound, got #<NoMethodError: undefined method `fb_systemd_reload' for cookbook: fb_systemd, recipe: reload :Chef::Recipe> with backtrace:
```

Then, per the comments, once you load fb_systemd, then it want modprobe.

Not sure I remember what network_scripts was for, but my guess is another error popped up after that.

jaymzh: Looks like whatever the dep was is gone. will close.

PR #183, opened 2021-11-30
Tune `def self.expand_filter` for slight perf increase

ChefAustin:
See: https://github.com/JuanitoFatas/fast-ruby#enumerablereverseeach-vs-enumerablereverse_each-code

### DIFF
--- a/cookbooks/fb_helpers/libraries/fb_helpers.rb
+++ b/cookbooks/fb_helpers/libraries/fb_helpers.rb
@@ -179,7 +179,7 @@ If the has is specified, it takes one or more of the following keys:
       array_filter.each do |f|
         keys = f.split('/')
         h = nil
-        keys.reverse.each do |k|
+        keys.reverse_each do |k|
           h = { k => h }
         end
         self.merge_hash!(hash_filter, h)


ChefAustin:
I think the CI is busted as I doubt my 1 character change spurred all these failures. 🤣

jaymzh:
Yeah our CI is fucked at the moment. Do you know what version of Ruby `reverse_each` was added in? Internal FB still supports some old-ish versions...

ChefAustin:
> Yeah our CI is fucked at the moment. Do you know what version of Ruby `reverse_each` was added in? Internal FB still supports some old-ish versions...

(Guessing by CI) Is Ruby v2.4 as far back as y'all need to support?

ChefAustin:
All good as far back as v1.8.6p287:

https://apidock.com/ruby/v1_8_6_287/Array/reverse_each

jaymzh:
Perfect, thanks for checking.

jaymzh (approved):
Thanks. One of the current employees will need to pull this in and merge it, and it's roadmapping week there, so it may be a few days. But definitely an easy win.

PR #184, opened 2021-12-01
Favor `respond_to?` over `begin...rescue`

ChefAustin:
We currently run `fasterer` in CI against our internal Chef repository and after updating our internal copy of `fb_helpers` (to be in-sync with this upstream) this `begin...rescue` block was flagged by `fasterer`:
```
fb_helpers/libraries/node_methods.rb:897 Don't rescue NoMethodError, rather check with respond_to?.
```

Implementing this small change could improve speed performance of the function _up to_ 8x.

I am open to alternative approaches to this but this was the first one that came to mind (and this repo already contains usages of `#respond_to?` so I didn't have to worry about compatibility of newly introduced methods and legacy Ruby versions in-use at Facebook).

See: https://github.com/JuanitoFatas/fast-ruby#beginrescue-vs-respond_to-for-control-flow-code

### DIFF
--- a/cookbooks/fb_helpers/libraries/node_methods.rb
+++ b/cookbooks/fb_helpers/libraries/node_methods.rb
@@ -892,9 +892,9 @@ class Chef
       # implicit-begin is a function of ruby2.5 and later, but we still
       # support 2.4, so.... until then
       node_path.inject(self) do |location, key|
-        begin
+        if key.respond_to?(:to_s) && location.respond_to?(:attribute?)
           location.attribute?(key.to_s) ? location[key] : default
-        rescue NoMethodError
+        else
           default
         end
       end


cookbooks/fb_helpers/libraries/node_methods.rb
```diff
@@ -892,9 +892,9 @@ def attr_lookup(path, delim: '/', default: nil)
       # implicit-begin is a function of ruby2.5 and later, but we still
       # support 2.4, so.... until then
       node_path.inject(self) do |location, key|
-        begin
+        if key.respond_to?(:to_s) && location.respond_to?(:attribute?)
```
jaymzh, line 895: This seems reasonable to me, but calling @davide125 who wrote this code to see if there was some magical reason he did `begin..rescue`.

davide125: I don't remember making a conscious decision on this (hell, I don't remember writing this at all lol), and I suspect this was mostly copypasta from somewhere else. I think this is fine, as long as the various tests in `fb_helpers` still pass.

jaymzh (approved):

cookbooks/fb_helpers/libraries/node_methods.rb
ChefAustin: > This seems reasonable to me, but calling @davide125 who wrote this code to see if there was some magical reason he did begin..rescue.


https://www.youtube.com/watch?v=XkfO8c8MlKU

> I don't remember making a conscious decision on this (hell, I don't remember writing this at all lol), and I suspect this was mostly copypasta from somewhere else. I think this is fine, as long as the various tests in fb_helpers still pass.

Awesome. Thanks for the taking a gander at this!

PR #185, opened 2021-12-03
Some lint fixes for fb_apache

jaymzh:
This handles newer rubocop/cookstyle fixes. Should still all
be compatible with Chef 12, though I believe you're all the way to 13.


davide125 (approved):

jaymzh:
There's other divergences but I'd like to get this simple one in to make the next one easier to review

PR #189, opened 2022-01-07
fb_iptables: Support custom user chains; minor bugfixes

jaymzh:
If a user defined a custom chain, `fb_iptables`, would try to set a
default policy of `nil`, which is invalid syntax (and, ironically, the
tests looked for this invalid syntax), and would never actually make the
chain (oops).

This PR does 2 things:

* Comments the template and fixes indentation so that it's readable
* Creates chains the user makes, but doesn't try to set default
policies, which aren't a thing on non-built-in-chains


cookbooks/fb_iptables/spec/fixtures/default/iptables_multi_chain
```diff
@@ -8,10 +8,10 @@
 COMMIT
 
 *filter
+-N LOG_DB
 :INPUT ACCEPT [0:0]
 :OUTPUT ACCEPT [0:0]
 :FORWARD ACCEPT [0:0]
-:LOG_DB  [0:0]
```
davide125, line 14: You sure this isn't actually needed?

davide125 (commented):
Overall logic lgtm

cookbooks/fb_iptables/spec/fixtures/default/iptables_multi_chain
jaymzh: It's actually a syntax error. It'll break it loading. You can, on some versions do `:LOG_DB - [0:0]`, but it's a no-op and never necessary. BUt it's a 3-column field we're generating 2 columns for.

davide125 (approved):

PR #190, opened 2022-01-10
fb_iptables: simple function to enable NAT

jaymzh:
FB doesn't have NAT in their kernels and thus fb_iptables doesn't
support it. It's easy enough to add support to it from a wrapper
cookbook by amending all the right data structures, but if two
different cookbooks do it they can step on each other.

This PR provides a method that others can call to setup the NAT table
structures safely/idempotently while still keeping it disabled by
default.

Talked this over with @dvaide125 and he agreed this was a mutually
agreeable approach.


davide125 (approved):

jaymzh:
This one is trivial, can I get someone to pull it in this week? It'll help with some other rebases and syncs CC @davide125 @scottruss @NaomiReeves

PR #194, opened 2022-02-04
Add overrides to the fb_dnsmasq API

jaymzh:
I discussed this with @davide125 way way back when, but somehow this
never got pushed up stream.

The unitfiles with systemd don't restart it when it does. That's silly,
fix that. Also give an API while we're at it to add additional pieces to
that override.

### DIFF
--- a/cookbooks/fb_dnsmasq/README.md
+++ b/cookbooks/fb_dnsmasq/README.md
@@ -10,6 +10,7 @@ Attributes
 ----------
 * node['fb_dnsmasq']['config']
 * node['fb_dnsmasq']['enable']
+* node['fb_dnsmasq']['systemd_overrides']
 
 Usage
 -----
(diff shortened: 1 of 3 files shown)


cookbooks/fb_dnsmasq/README.md
```diff
@@ -44,3 +45,14 @@ the `fb_hosts` cookbook.
 If the `read-ethers` config option is set, dnsmasq will read mac address to
 hostname mappings from `/etc/ethers`. This can be customized using the API
 provided by the `fb_ethers` cookbook.
+
+### systemd_overrides
+You can set any systemd_overrides in `node['fb_dnsmasq']['systemd_overrides']`,
+and they will use the `fb_systemd_override` resource to create a drop-in file
+for systemd. The default in this cookbook turns Restart=always on for the
+service since some packges don't set it at all. Here's how you would set that
```
joshuamiller01, line 53: packages

cookbooks/fb_dnsmasq/README.md
```diff
@@ -44,3 +45,14 @@ the `fb_hosts` cookbook.
...
+same thing explicity:
```
joshuamiller01, line 54: explicitly

jaymzh:
typos fixed

PR #201, opened 2022-02-08
Fix fb_logrotate craziness

jaymzh:
In
https://github.com/facebook/chef-cookbooks/commit/c5104612bf77d35474cff57e4c85ca000d848f53,
@malmond made the faulty claim that `whyrun_safe_ruby_block` was
dangerous, but he misunderstood chef ordering. It's the very very last
thing before the template (except the validator), and no one can
possibly schedule things after it, but before the template. It's the
whole reason why `whyrun_safe_ruby_block` exists.

Further, debugging templates is difficult. And logic doesn't belong in
templates. And certainly not munging of configs.

This moves the logic back to where it belongs, being careful to also
keep the logic fixes from that PR.

### DIFF
--- a/cookbooks/fb_logrotate/templates/default/fb_logrotate.conf.erb
+++ b/cookbooks/fb_logrotate/templates/default/fb_logrotate.conf.erb
@@ -25,42 +25,11 @@ size  <%= node['fb_logrotate']['globals']['size'] %>
 compressext <%= node['fb_logrotate']['globals']['compressext'] %>
 <% end -%>
 
-<% node['fb_logrotate']['configs'].to_hash.each do |name, block|
-  time = nil
-  if block['overrides']
-    rotation = block['overrides']['rotation']
-    size = block['overrides']['size']
-    if rotation
-      if size
-        fail "fb_logrotate:[#{name}]: you can only use size or rotation not both"
-      end
-      if rotation == 'weekly' && !block['overrides']['rotate']
-         block['overrides']['rotate'] = '4'
-      end
-      if %w{hourly daily weekly monthly yearly}.include?(rotation)
-        time = rotation
-        block['overrides'].delete('rotation')
-      else
-        fail "fb_logrotate:[#{name}]: rotation #{rotation} invalid"
-      end
-    end
-    if size
-      time = "size #{size}"
-      block['overrides'].delete('size')
-    end      
-    if time == defaultrotation
-      time = nil
-    end
-    if block['overrides']['nocompress'] &&
-      node['fb_logrotate']['globals']['nocompress']
-      # redundant, remove.
-      block['overrides'].delete('nocompress')
-    end
-  end -%>
+<% node['fb_logrotate']['configs'].to_hash.each do |name, block| %>
 ## fb.fb_logrotate.configs.<%= name %>
 <%=  block['files'].join(' ') %> {
-<%   if time -%>
-  <%= time %>
+<%   if block['time'] -%>
+  <%=  block['time'] %>
 <%   end -%>
 <%   if block['overrides'] -%>
 <%     block['overrides'].keys.sort.each do |k| -%>
--- a/cookbooks/fb_logrotate/README.md
+++ b/cookbooks/fb_logrotate/README.md
@@ -32,7 +32,7 @@ per logrotate configuration block.
 
 To rotate a new set of logs, add a new entry to this hash, like so:
 
-```
+```ruby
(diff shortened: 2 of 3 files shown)


cookbooks/fb_logrotate/templates/default/fb_logrotate.conf.erb
```diff
@@ -25,42 +25,11 @@ size  <%= node['fb_logrotate']['globals']['size'] %>
...
 <% end -%>
 
-<% node['fb_logrotate']['configs'].to_hash.each do |name, block|
-  time = nil
-  if block['overrides']
-    rotation = block['overrides']['rotation']
-    size = block['overrides']['size']
-    if rotation
-      if size
-        fail "fb_logrotate:[#{name}]: you can only use size or rotation not both"
```
joshuamiller01, line 35: Why drop this?

cookbooks/fb_logrotate/templates/default/fb_logrotate.conf.erb
```diff
@@ -25,42 +25,11 @@ size  <%= node['fb_logrotate']['globals']['size'] %>
...
-      end
-      if rotation == 'weekly' && !block['overrides']['rotate']
-         block['overrides']['rotate'] = '4'
-      end
-      if %w{hourly daily weekly monthly yearly}.include?(rotation)
-        time = rotation
-        block['overrides'].delete('rotation')
-      else
-        fail "fb_logrotate:[#{name}]: rotation #{rotation} invalid"
```
joshuamiller01, line 44: And this?

cookbooks/fb_logrotate/templates/default/fb_logrotate.conf.erb
jaymzh: Oversight, will fix, thanks!

davide125 (changes requested):
This breaks if no rotation override:
```[2022-02-17T13:57:21-08:00] INFO: Processing whyrun_safe_ruby_block[munge logrotate configs] action run (fb_logrotate::default line 35)

================================================================================
Error executing action `run` on resource 'whyrun_safe_ruby_block[munge logrotate configs]'
================================================================================

RuntimeError
------------
fb_logrotate:[syslog]: rotation  invalid

Cookbook Trace:
---------------
/etc/chef/local-mode-cache/cache/cookbooks/fb_logrotate/recipes/default.rb:57:in `block (3 levels) in from_file'
/etc/chef/local-mode-cache/cache/cookbooks/fb_logrotate/recipes/default.rb:37:in `each'
/etc/chef/local-mode-cache/cache/cookbooks/fb_logrotate/recipes/default.rb:37:in `block (2 levels) in from_file'

Resource Declaration:
---------------------
# In /etc/chef/local-mode-cache/cache/cookbooks/fb_logrotate/recipes/default.rb

 35: whyrun_safe_ruby_block 'munge logrotate configs' do
 36:   block do
 37:     node['fb_logrotate']['configs'].to_hash.each do |name, block|
 38:       time = nil
 39:       if block['overrides']
 40:         rotation = block['overrides']['rotation']
 41:         size = block['overrides']['size']
 42:         if rotation && size
 43:           fail "fb_logrotate:[#{name}]: you can only use size or rotation " +
 44:             'not both'
 45:         end
 46:         # if someone wants to override weekly but didn't specify
 47:         # how many to keep, we default to 4
 48:         if rotation == 'weekly' && !block['overrides']['rotate']
 49:           node.default['fb_logrotate']['configs'][name][
 50:             'overrides']['rotate'] = '4'
 51:         end
 52: 
 53:         if %w{hourly daily weekly monthly yearly}.include?(rotation)
 54:           time = rotation
 55:           node.rm(:fb_logrotate, :configs, name, :overrides, :rotation)
 56:         else
 57:           fail "fb_logrotate:[#{name}]: rotation #{rotation} invalid"
 58:         end
 59: 
```
with
```
chef (13.10.0)> pp node['fb_logrotate']['configs']['syslog']
{"files"=>
  ["/var/log/cron",
   "/var/log/maillog",
   "/var/log/messages",
   "/var/log/secure",
   "/var/log/spooler"],
 "overrides"=>
  {"nocreate"=>true,
   "nocopytruncate"=>true,
   "missingok"=>true,
   "sharedscripts"=>true,
   "postrotate"=>"/bin/systemctl kill -s HUP rsyslog || true"}}
```

jaymzh:
Good call. Fixed.

PR #203, opened 2022-02-09
new cookbook: fb_smartmon

jaymzh:
This is a very simple cookbook to install and configure smartmontools.

This was primarily written by @infcted


davide125:
Can you contribute to https://github.com/facebook/chef-cookbooks/tree/main/cookbooks/fb_smartctl instead? Looks like it's roughly the same scope.

PR #208, opened 2022-02-10
Lint fb_storage; add mdadm package

jaymzh:
Syncing fb_storage with our internal fixes/changes, small PR, autofix
lint and mdadm package support


cookbooks/fb_syslog/recipes/default.rb
```diff
@@ -73,6 +73,12 @@
   notifies :restart, "service[#{service_name}]"
 end
 
+directory '/etc/rsyslog.d' do
+  action :delete
+  recursive true
+  notifies :restart, "service[#{service_name}]"
+end
+
```
davide125, line 81: Can you spin this out in a separate PR ? I just checked and we still have a bunch of crap in there so it'd need a careful rollout.

davide125 (changes requested):

cookbooks/fb_syslog/recipes/default.rb
davide125: Alternatively, gate it behind an attribute knob

cookbooks/fb_syslog/attributes/default.rb
```diff
@@ -43,9 +43,14 @@
...
       'selector' => '*.info;mail,authpriv,cron,' +
-        'local0,local1,local2,local3,local5,local6,local7.none',
+        'local0,local1,local2,local3,local4,local5,local6,local7.none',
       'action' => "-#{syslog_file}",
     },
+    'authlog' => {
+      'comment' => 'Log all auth stuff',
+      'selector' => 'auth,authpriv.*',
+      'action' => '/var/log/auth.log',
+    },
```
davide125, line 53: This is a Debian thing, so it should probably be gated to Debian/Ubuntu, CentOS and friends use /var/log/secure for this stuff.

jaymzh: well it should probably just have the right name on RHEL and friends. Not having the auth/secure logs is a bummer. Will do a conditional.

PR #218, opened 2023-03-28
fix(fb_apache): render arrays inside of hash configs

ericnorris:
## Description

This PR fixes a bug where the `template_hash_handler` in `fb_apache` silently skips over `Array` values. Now these values are rendered as if they were in the top-level `VirtualHost` directive.

For example:

```ruby
{
  ...
  'Location /foo' => {
    'Header' => [
      'unset Bar',
      'unset Baz',
    ],
    ...
```

...will now properly include multiple `Header ...` lines in the `Location /foo` directive.

## Impact

These values were ignored previously, so it is possible that it would cause changes in hosts where this was happening.


jaymzh (changes requested):
Thanks for contributing!

This is interesting. It doesn't handle hashes because of https://github.com/facebook/chef-cookbooks/blob/main/cookbooks/fb_apache/templates/default/apache_conf.erb#L4 where we handle arrays. However, if there's an array embedded in a Hash, you're right, we don't handle it.

So, let's DRY this up a bit. Update `apache_conf.erb` to call this function if it's array *or* a hash, and rename the method to `template_data_handler` (or if you have a better name, I'm open).

And finally, you'll need to sign the CLA.

ericnorris:
Hey @jaymzh, thanks for the quick review! I hope you don't mind but I went a step further and moved all of the template logic into the function. This had the added benefit of allowing the `_rewrites` config in both the virtual host and directory contexts; if I understand correctly it could only be inside a directory context previously.

jaymzh:
Can you even have rewrites inside a directory? I thought they only worked inside a VirtualHost

ericnorris:
I believe so! I tested this locally, and per https://httpd.apache.org/docs/current/mod/mod_rewrite.html#rewriterule:

> Context: server config, virtual host, directory, .htaccess

Also, if I understand correctly my change actually enables this for _virtual hosts_, whereas before it was only allowed inside a directory. This is because the top-level `apache_conf.erb` didn't handle the `_rewrite` block, so it could only be called when you had specified a `Location` or `Directory` sub-hash.

jaymzh:
Yeah I was gonna say, it definitely worked at the top level, we use it at SCALE:

https://github.com/socallinuxexpo/scale-chef/blob/main/cookbooks/scale_apache/recipes/default.rb#L203

jaymzh:
This looks reasonable, but you need to rebase. I kicked off the tests... I know a lot are broken (sorry), but lets make sure any Apache ones pass.

ericnorris:
@jaymzh I'm still waiting on getting the CLA signed, apologies for that.

That said, I've rebased the branch and fixed a `rubocop` lint check (the copy-pasted logic was looking at `Fixnum` instead of `Integer`, which was caught by the lint). All of the other failing tests appear unrelated to my change, while there is something about `fb_apache` in the `ruby (2.6)` check, I've seen that failing in other builds (e.g. the current commit on `main`: https://github.com/facebook/chef-cookbooks/actions/runs/4640529922/jobs/8212557841).

jaymzh (approved):
LGTM. Once you've signed the CLA, someone at Meta can pull this in internally, test it, and merge it.

ericnorris:
I'm still stuck on trying to get the CLA signed; we already had an existing CLA and we've sent an email to cla@meta.com to update it but have not yet received a response. We're interested in potentially submitting other PRs in the future so we're motivated to figure this out.

jaymzh:
Hey @davide125 or @joshuamiller01 - maybe one of you can poke the opensource folks on the status of Eric's request?

dafyddcrosby:
> I'm still stuck on trying to get the CLA signed; we already had an existing CLA and we've sent an email to [cla@meta.com](mailto:cla@meta.com) to update it but have not yet received a response. We're interested in potentially submitting other PRs in the future so we're motivated to figure this out.

Taking a look into the CLA issue now - did you try https://code.facebook.com/cla as well?

ericnorris:
Thanks @jaymzh and @dafyddcrosby! I was considering signing a separate CLA, but it seems that someone responded soon after I left the earlier GitHub message. I believe my coworkers and myself are now squared away with the CLA, although I see the check hasn't yet updated.

ericnorris:
@jaymzh no worries! I'm not sure if you meant a literal rebase, but I was able to merge the latest commits via the GitHub UI. If you want a literal rebase just let me know and I can get around to it.

jaymzh:
That's fine. We tend to use a rebase-based workflow, but it'll get sorted out int he squash-merge anyway. GH displayed it correctly so that's fine.

PR #219, opened 2023-04-07
fb_smokeping: fix debian packages list

bwann:
Metadata claims it supports Debian and Ubuntu but this was breaking tests because Debian packages weren't provided


cookbooks/fb_smokeping/recipes/default.rb
```diff
@@ -19,7 +19,7 @@
 #
 
 packages = value_for_platform(
-  'ubuntu' => { :default => %w{fcgiwrap smokeping} },
+  ['debian', 'ubuntu'] => { :default => %w{fcgiwrap smokeping} },
 )
```
jaymzh, line 24: I would also do something like:

```
unless packages
  die 'fb_smokeping included on an unsupported distro.'
end

jaymzh (changes requested):

PR #224, opened 2023-07-11
fix(fb_modprobe, fb_systemd): updates to recipes for ubuntu jammy

PatrickDale:
## Description

`fb_modprobe`
- Updates the cookbook metadata to support `ubuntu`.

`fb_systemd`
- Adds the required `systemd-timesyncd` package when `timesyncd` is enabled.
- Adds the `dns` resolver to the `fb_nsswitch` configuration to be used as a fallback to `systemd-resolved`. This is a recommended configuration for `nss-resolve`: https://man7.org/linux/man-pages/man8/nss-resolve.8.html.

## Impact

These cookbooks will now be able to converge on Ubuntu Jammy.


cookbooks/fb_modprobe/metadata.rb
```diff
@@ -7,5 +7,6 @@
 source_url 'https://github.com/facebook/chef-cookbooks/'
 version '0.0.1'
 supports 'centos'
+supports 'ubuntu'
```
jaymzh, line 10: lets add debian while we're at it.

cookbooks/fb_systemd/recipes/default_packages.rb
```diff
@@ -35,7 +35,7 @@
   end
 
   # older versions of Debian and Ubuntu are missing some extra packages
-  unless ['trusty', 'jessie'].include?(node['lsb']['codename'])
+  if (node.ubuntu? && node['platform_version'].to_i < 14) || (node.debian? && node['platform_version'].to_i < 8)
```
jaymzh, line 38:
```suggestion
  if (node.ubuntu? && node['platform_version'].to_i < 14) ||
     (node.debian? && node['platform_version'].to_i < 8)
```

cookbooks/fb_systemd/recipes/default_packages.rb
```diff
@@ -59,6 +59,9 @@
        !['trusty', 'jessie'].include?(node['lsb']['codename'])
       systemd_packages << 'systemd-journal-remote'
     end
+    if node['fb_systemd']['timesyncd']['enable'] && node.debian?
```
jaymzh, line 62: this is not runtime safe, you cannot read the node at compile time, no one will have had a chance to set it yet.

We should just always install it.

We COULD make a separate package resource and use an `only_if` here, if there's a good reason, but our default is to install all the relevant software unless `manage_packages` is `false`.

cookbooks/fb_systemd/recipes/resolved.rb
```diff
@@ -31,10 +31,11 @@
...
 
 # nss-resolve enables DNS resolution via the systemd-resolved DNS/LLMNR caching
-# stub resolver. According to upstream this should replace the glibc "dns"
-# resolver and is required for systemd-resolved to work. This block attempts
-# to place the resolver between mymachines and myhostname as recommended by
+# stub resolver. According to upstream this should be used in favor of the glibc
+# "dns" resolver and is required for systemd-resolved to work. This block attempts
+# to place "resolver dns" between mymachines and myhostname as recommended by
 # upstream.
+# See: https://man7.org/linux/man-pages/man8/nss-resolve.8.html
```
jaymzh, line 38: Calling @davide125 and @anitazha to aisle 3...

I think this is probably fine, but there may be a performance or other reason to avoid that. If so, we should make it configurable.

jaymzh (changes requested):

cookbooks/fb_systemd/recipes/default_packages.rb
PatrickDale: I updated this to always install timesyncd if it is enabled on all platforms:
```
if node['fb_systemd']['timesyncd']['enable']
```

I believe it is safe to read this node attribute since there is a `lazy` block wrapping everything: https://github.com/facebook/chef-cookbooks/blob/main/cookbooks/fb_systemd/recipes/default_packages.rb#L56

but I'm happy to move this package elsewhere in this file! it felt funny always installing it since someone could not choose to disable `timesyncd`.

PatrickDale: thanks for the speedy feedback on this PR btw!

cookbooks/fb_systemd/recipes/default_packages.rb
```diff
@@ -59,6 +60,9 @@
        !['trusty', 'jessie'].include?(node['lsb']['codename'])
       systemd_packages << 'systemd-journal-remote'
     end
+    if node['fb_systemd']['timesyncd']['enable']
```
jaymzh, line 63: Oh you're right, sorry I didn't expand out the diff. This is totally good.

jaymzh (commented):
Thanks, this looks good. I'll ping Davide offline to make sure he saw my tag here so we can get a final answer on that and unblock you.

cookbooks/fb_systemd/recipes/default_packages.rb
PatrickDale: thanks!

PatrickDale:
> Thanks, this looks good. I'll ping Davide offline to make sure he saw my tag here so we can get a final answer on that and unblock you.

Sounds great, thanks again for the review!

cookbooks/fb_systemd/recipes/resolved.rb
davide125: I did some spelunking and historically `resolve` was supposed to replace `dns`, which is why this was forcibly removing it. It looks like that guidance has now changed so inserting it before `dns` should be fine. I'm not in love with this whole logic as it's pretty brittle, but reworking it and adding some tests is probably something for another time.

PR #225, opened 2023-07-21
feat(postfix): add support for ubuntu

PatrickDale:
## Description

- Allows `fb_postfix` to be used on Debian families.
- Removes default `fb_postfix` `main.cf` settings for Debian families since we see errors from them.

## Context / Why are we making this change?

To use `fb_postfix` on Ubuntu.


cookbooks/fb_postfix/attributes/default.rb
```diff
@@ -15,18 +15,11 @@
...
-  'enable' => true,
-  'mask_service' => false,
-  'aliases' => {},
-  'access' => {},
-  'canonical' => {},
-  'etrn_access' => {
-    '127.0.0.1' => 'OK',
+main_cf = value_for_platform_family(
+  'debian' => {
+    'compatibility_level' => '3.6',
```
mrichins, line 20: Am I right in thinking we would need most of the `default` config as well

PatrickDale: I was able to send mail using the default `postfix` settings just fine!

jaymzh (changes requested):
I'm currently using this cookbook on Debian and it works fine....

jaymzh (changes requested):
Just to follow up here with a few more details.

I'm fine bumping the compatibility level - we can probably do that everywhere, frankly.

Dropping most of the default config doesn't seem appropriate, most of these configs are the defaults from main.conf.proto - similar across both Debian and Fedora. If there's problematic ones, happy to nuke them.

Just FYI I'm using this on Debian sid without an issue.

jaymzh:
I'm gonna go ahead and close this for now - but if you have further comments or questions, feel free to re-open.

PR #229, opened 2023-08-21
[github] Update CI to Chef 16, drop Ruby 2.5/2.6 support

dafyddcrosby:
The end result of this is working rubocop/chefspec tests (which have been broken for a while now, so I'm happy to be seeing green). Test Kitchen fixes are out of scope for this one, since there's a bunch more work in that area.


jaymzh (approved):

jaymzh:
nice, thanks!

PR #232, opened 2023-12-28
fb_apt: Add ability to log output of `apt-get update`

adsr:
We're using `Debug::pkgAcquire::Worker` and this patch to debug https://bugs.launchpad.net/ubuntu/+source/apt/+bug/2003851 which we run into occasionally.


joshuamiller01:
Huh, I ... somehow broke the workflow I think.

joshuamiller01:
Gonna just fix this internally and push the fix, hopefully that'll work.

joshuamiller01:
I can't import for whatever reason, but I manually fixed this, should be good.

PR #235, opened 2024-02-22
fb_postfix: default to lmdb maps for el10 and later

davide125:
Summary:
Upstream dropped support for libdb in
https://src.fedoraproject.org/rpms/postfix/c/6a2621e4d73d59337dc64ba45922132286b841a1
so we need to use lmdb in RHEL 10 and later

Differential Revision: D54077466

### DIFF
--- a/cookbooks/fb_postfix/attributes/default.rb
+++ b/cookbooks/fb_postfix/attributes/default.rb
@@ -15,6 +15,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# RHEL 10 and later don't support hash maps anymore
+# https://src.fedoraproject.org/rpms/postfix/c/6a2621e4d73d59337dc64ba45922132286b841a1
+if node.el_min_version?(10) || node.eln?
+  map_type = 'lmdb'
+else
+  map_type = 'hash'
+end
+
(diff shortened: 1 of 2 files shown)


davide125:
iproute commit is unrelated, I didn't mean to export it in this PR

cookbooks/fb_postfix/attributes/default.rb
```diff
@@ -15,6 +15,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# RHEL 10 and later don't support hash maps anymore
+# https://src.fedoraproject.org/rpms/postfix/c/6a2621e4d73d59337dc64ba45922132286b841a1
+if node.centos_min_version?(10) || node.eln?
+  map_type = 'lmdb'
+else
+  map_type = 'hash'
+end
```
jaymzh, line 24: shouldn't this just be el_min_version? this applies to RHEL too, right?

jaymzh (changes requested):

cookbooks/fb_postfix/attributes/default.rb
davide125: Yeah, good call.

PR #236, opened 2024-03-15
Fix fb_systemd on hosts without networkd

jaymzh:
Fix fb_systemd on hosts without networkd

Summary:

If you don't use networkd (and `fb_networkd`), then this code in `fb_system::networkd`

Test Plan:

@anitazha did some testing of this internal to Meta already, plus
I've rolled this out to the SCALE infra


dafyddcrosby (approved):

dafyddcrosby:
Looks like this didn't get auto-closed, thanks again 👍

PR #237, opened 2024-04-11
Remove supermarket_push script

dafyddcrosby:
I didn't see any of the cookbooks on the supermarket, suspect the script is derelict.
@jaymzh is this still in use on your end?


jaymzh:
They broke our ability to push to supermarket, you can no longer push the same version repeatedly, so it's unfortunate, but yeah. I think it'd be great to push them.

I haven't tried in a long time though.

PR #239, opened 2024-05-23
Update checkout GHA to avoid Node16 deprecation warnings

furan917:
This is a very basic PR I know. But it cleans up the deprecation warnings so I thought I'd add it.


jaymzh (approved):

jaymzh:
@joshuamiller01 - I don't think these files are controlled by `shipit`, but since I can't confirm that, just in case I won't merge and will leave to you.

dafyddcrosby (approved):
Thanks for submitting!

PR #240, opened 2024-07-29
Explicitly depend on fb_sysfs in fb_storage

russtaylor:
I spun up a new class of GCE VM which uses NVMe-based storage, but fb_storage failed, complaining about being unable to find fb_sysfs. After digging into it a bit, I noticed that fb_storage doesn't explicitly depend on `fb_sysfs`, despite using it in several places, [like when configuring nvme-based storage](https://github.com/facebook/chef-cookbooks/blob/c64381a2e0195855703f59a001a64a589ae1ac73/cookbooks/fb_storage/resources/format_devices.rb#L169).

I verified that simply adding `depends 'fb_sysfs'` was enough to resolve the error.


jaymzh (approved):

jaymzh:
LGTM - I'll wait for someone internal to pull it internally.

Issue #233, opened 2024-01-02
Compound API interactions for cookbooks included by `fb_init` are difficult to implement


ericnorris:
# Context

At Etsy, we've gone all-in on using this repository and the philosophy it represents (which we call the "Facebook pattern"), and have been working on refactoring our Chef code repository in a greenfield project.

As such, we've written an `etsy_init` (i.e. `fb_init`) cookbook that follows the same pattern as [`fb_init_sample`](https://github.com/facebook/chef-cookbooks/tree/main/cookbooks/fb_init_sample) - we have a bunch of settings we configure for our "base", and then a number of `include_recipe 'fb_foo'` statements to match.

We also apply our cookbooks with the run list as suggested by the README - [`recipe[fb_init], recipe[$SERVICE]`](https://github.com/facebook/chef-cookbooks/tree/e17c84060949f3a6fde4da8dd508d1492267402f#run-lists) - where in our case `$SERVICE` is `hostgroup_foo`, i.e. a cookbook that configures everything necessary for a `foo` cluster.

These "hostgroup" cookbooks often configure and include a cookbook that wraps some "Facebook-style" (our term) cookbooks, like `fb_apache`, into a reusable building block that can be shared by nearly similar hostgroups. We refer to these as `stack` cookbooks, but the naming isn't particularly important - what is important is that these cookbooks will almost always have "[compound API interactions](https://github.com/facebook/chef-utils/blob/88740584169cba30dad68c69485cfa969838b12a/Compile-Time-Run-Time.md#compound-api-interactions)".

To make this more concrete, as an example we may have a `stack_etsyweb` cookbook that:

- Sets up `fb_apache`
- Installs php (imagine an `fb_php` cookbook)
- Configures crons
- Etc.

...all for running "Etsyweb", our monolithic PHP application.

While the application itself is a monolith, we do run the application on separate clusters, and these clusters may be configured slightly differently. Due to this, our `stack_etsyweb` cookbook will expose an API that a `hostgroup_foo` can configure differently than a `hostgroup_bar` would, but otherwise they are exactly the same.

The API exposed by `stack_etsyweb` is, as mentioned above, a [compound API interaction](https://github.com/facebook/chef-utils/blob/88740584169cba30dad68c69485cfa969838b12a/Compile-Time-Run-Time.md#compound-api-interactions), and so it will set `fb` attributes at runtime, like so:

```ruby
whyrun_safe_ruby_block 'foo' do
  block do
    # ...
  end
end
[...]

PR #242, opened 2024-11-08
feat: lazy timer commands

ericnorris:
Before this commit, it was not possible to lazily declare an `fb_timers` entry that utilized a node attribute inside the command.

After this commit, specifying a `Proc` for `command` or entries in `commands` will cause `fb_timers` to evaluate the proc when rendering the service template.

This is related to https://github.com/facebook/chef-cookbooks/issues/233. I can't exactly recall where we landed on whether it was acceptable to add `Proc` support to cookbooks, but this particular example has come up enough times that I feel that it's worth enabling.

(I couldn't run these tests locally, so I might need to iterate a bit if I made a mistake)

### DIFF
--- a/cookbooks/fb_timers/README.md
+++ b/cookbooks/fb_timers/README.md
@@ -65,7 +65,8 @@ Required fields:
   when you want your job to run. Corresponds to the `OnCalendar` field of the
   systemd timer. See below for helpers to generate common calendar patterns.
 * `command`: The command to run. Corresponds to the `ExecStart` field of the
-  systemd service.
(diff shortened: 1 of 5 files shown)


jaymzh (approved):
Yes, this is definitely the preferred way to do these things.

ericnorris:
Thanks @jaymzh! I think the `kitchen` test failures are unrelated? I believe the spec tests look good though.

jaymzh:
yeah kitchen tests have been broken a while, and no one has had time to go fix them.

@joshuamiller01 or @dafyddcrosby or someone will pull this in internally to run against extra internal tests and then merge if there's no issues.

PR #245, opened 2024-11-14
Fix fb_apache for Debian-family

jaymzh:
The `verify_configs` custom resource wasn't using the passed-in httpdir
and was assuming RHEL-like httpdir and config file name. Fixed both.

Also, on Debian sid, the version-matching failed, so fixed that.


dafyddcrosby:
Not sure why this wasn't auto-closed, but this has now been merged 👍

jaymzh:
Thanks!

PR #250, opened 2025-02-03
fb_apt: Significant updates

jaymzh:
* Deprecate `node['fb_apt']['repos']` which was always a bad API
  (sorry), and replace it with `node['fb_apt']['sources']` which
  integrates nicely with the new `node['fb_apt']['keymap']`
* Deprecate `node['fb_apt']['keys']` which was very broken on modern
  apt and replace it with a new `node['fb_apt']['keymap']`
* Update syntax for security and update repos on modern debian and
  ubuntu
* Remove old Ubuntu 16 cruft
* Lots of cleanups and refactoring for readability

Signed-off-by: Phil Dibowitz <phil@ipom.com>

PR #254, opened 2025-02-06
 CI fixes: yum repos; disable logind; chef 18

jaymzh:
* centos8 mirrors have moved, adjust kitchen setup accordingly.
* logind can't run in containers
* modern chef
* Newer chef requires at least debian 11 for SSL compatability
* rsyslog needs several tweaks to run in containers, added to ci_fixes
* default apache config in centos points to some certs, so make those

This makes everything green _except_ for debian which requires
significant refactors of fb_apt, which can be found in #250, but I 
didn't want production-effecting stuff mixed with CI fixes

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,7 +30,7 @@ jobs:
         os:
           - centos-stream-8
           - ubuntu-2004
-          - debian-10
+          - debian-11
     runs-on: ubuntu-latest
     steps:
     - name: Checkout repository
@@ -47,7 +47,7 @@ jobs:
         os: ${{ matrix.os }}
       env:
         CHEF_LICENSE: accept-no-persist
-        CHEF_VERSION: 16.18.0
+        CHEF_VERSION: 18.6.14
   shellcheck:
     runs-on: ubuntu-latest
     steps:
--- a/cookbooks/fb_syslog/recipes/default.rb
+++ b/cookbooks/fb_syslog/recipes/default.rb
@@ -77,6 +77,6 @@ service service_name do
   action :start
   subscribes :restart, 'package[rsyslog]'
   # within vagrant, sometimes rsyslog fails to restart the first time
-  retries 5
-  retry_delay 5
+  retries 1
+  retry_delay 15
 end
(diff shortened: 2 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -47,7 +47,7 @@ jobs:
         os: ${{ matrix.os }}
       env:
         CHEF_LICENSE: accept-no-persist
-        CHEF_VERSION: 16.18.0
+        CHEF_VERSION: 17.10.163
```
dafyddcrosby, line 50: Mind changing that to something more recent, say 18.5.0?

dafyddcrosby (changes requested):
Let's use the 18 branch, since that's what we've been contributing upstream with

.github/workflows/ci.yml
jaymzh: Done

dafyddcrosby (approved):

dafyddcrosby:
Crap, I forgot that there was that Docker publishing issue with 18.5.0.  https://hub.docker.com/r/chef/chef/tags?name=18.

Let's say 18.6.14 instead? https://hub.docker.com/layers/chef/chef/18.6.14/images/sha256-a303c5226a51f02108de1c3bd9cc99805de81de5d0c32455fb3aa30b9eda547f We've backported a lot of that stuff, so it's a close enough approximation for now

cookbooks/fb_syslog/recipes/default.rb
```diff
@@ -77,6 +77,7 @@
   action :start
   subscribes :restart, 'package[rsyslog]'
   # within vagrant, sometimes rsyslog fails to restart the first time
-  retries 5
-  retry_delay 5
+  retries 1
+  retry_delay 15
+  ignore_failure true
```
davide125, line 82: Can we somehow detect the vagrant test environment instead of doing this?

jaymzh: the ignore_failure shouldn't be there, that was from testing. I will remove. As to the vagrant stuff I suspect it's not needed anymore and dropped the retries down to only one, but didn't want to remove them altogether in case firstboot_os internally had somehow come to rely on them

PR #255, opened 2025-02-06
fix tests: remove fb_users from kitchen

jaymzh:
fb_users requires users to fill in UID_MAP, which we haven't done
for the kitchen VMs, so it doesn't make sense to run it in kitchen,
or you just get errors. So remove it for now.

Kitchen tests don't pass, but they get further now and I'm trying to do one -fix-per-PR rather than some huge PR.


dafyddcrosby (approved):

dafyddcrosby:
Quick note, Cookstyle failing on the commented out line

jaymzh:
Thanks, fixed

dafyddcrosby:
Fixed with ac723c51aa140523656dcb5a9c8405431f086b6c

PR #256, opened 2025-02-06
Fix typo in kitchen.yml

jaymzh:
kitchen.yml had a typo that was preventing debian containers from
using systemd which was causing a lot of failures.

This is based on top of #255 - and together Debian kitchen tests
now pass.

### DIFF
--- a/cookbooks/fb_init_sample/recipes/default.rb
+++ b/cookbooks/fb_init_sample/recipes/default.rb
@@ -79,7 +79,9 @@ include_recipe 'fb_limits'
 include_recipe 'fb_hostconf'
 include_recipe 'fb_sysctl'
 # HERE: networking
-include_recipe 'fb_users'
+# until we defined a UID_MAP that works with testing, this can't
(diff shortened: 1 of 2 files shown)


cookbooks/fb_init_sample/recipes/default.rb
```diff
@@ -79,7 +79,9 @@
 include_recipe 'fb_hostconf'
 include_recipe 'fb_sysctl'
 # HERE: networking
-include_recipe 'fb_users'
```
dafyddcrosby, line 82: Looks like this got pulled in by mistake

jaymzh: not a mistake. As I said in the PR description:
```
This is based on top of https://github.com/facebook/chef-cookbooks/pull/255 - and together Debian kitchen tests
now pass.
```

I can't test this without that as it never gets there. Once you merge the users one I"ll rebase all these

PR #257, opened 2025-02-06
Disable logind in kitchen

jaymzh:
It is masked on containers, so it breaks the run.


jaymzh:
Rolled this into another PR.

PR #279, opened 2025-02-19
Rework the `*_{min,max}_version?` functions; fix `fb_systemd`

jaymzh:
* Rework these functions to both be back-compat but also
have a mode that will work as desired.
* Add a lot of tests
* update fb_systemd to use real functions that exist in the real world

### DIFF
--- a/cookbooks/fb_networkmanager/resources/system_connections.rb
+++ b/cookbooks/fb_networkmanager/resources/system_connections.rb
@@ -29,6 +29,10 @@ action :manage do
     files = determine_files(name, config)
 
     current, new_config = generate_config_hashes(files['from'], config)
+    # Give it the friendly name the user wanted
+    unless new_config['id']
+      new_config['connection']['id'] = name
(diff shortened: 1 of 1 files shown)


cookbooks/fb_helpers/libraries/node_methods.rb
```diff
@@ -110,12 +127,12 @@ def rhel10?
       self.rhel? && self['platform_version'].start_with?('10')
     end
 
-    def centos_min_version?(version)
-      self.centos? && self.el_min_version?(version)
+    def centos_min_version?(version, _full = false)
```
joshuamiller01, line 130: What's the deal with not handling `full` here (and with oracle below)?  Are these platforms inherently never going to have non-integer versions?

jaymzh: A typo, supposed to be passing in full here. Will fix, good catch!

PR #283, opened 2025-02-23
fb_systemd: don't enable resolved when not enabled

jaymzh:
`libnss-resolve` pulls in `systemd-resolved`, which overwrites
`/etc/resolv.conf` on installation, but if you haven't enabled
systemd-resolved, then it's down and everything breaks.

So don't install it unless we're also planning to install resolved.


jaymzh:
This was closed in https://github.com/facebook/chef-cookbooks/commit/1f7a17013aa301292061f7bbe01811ca4ee83509

PR #287, opened 2025-02-26
Some fb_postfix fixups for Debian/Ubuntu

jaymzh:
There are a few of our defaults that weren't correct on Debian/Ubuntu. I
suspect these are the root of what @PatrickDale was running into in #225.
I didn't see it because my local cookbooks overwrote these three
variables.

However, I'm rolling this out to more Debian systems and bumped into it,
so fixing it.

### DIFF
--- a/cookbooks/fb_postfix/attributes/default.rb
+++ b/cookbooks/fb_postfix/attributes/default.rb
@@ -23,6 +23,16 @@ else
   map_type = 'hash'
 end
 
+if ::ChefUtils.debian?
+  daemon_directory = '/usr/lib/postfix/sbin'
+  mailq_path = '/usr/bin/mailq'
+  mailbox_command = nil
+else
+  daemon_directory = '/usr/libexec/postfix'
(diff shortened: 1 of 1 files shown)


cookbooks/fb_postfix/attributes/default.rb
```diff
@@ -68,7 +78,7 @@
     'sendmail_path' => '/usr/sbin/sendmail.postfix',
     'setgid_group' => 'postdrop',
     'manpage_directory' => '/usr/share/man',
-    'mailq_path' => '/usr/bin/mailq.postfix',
+    'mailq_path' => mailq_path,
```
davide125, line 81: This changes the default path on non-Debian platforms, is that intentional?

davide125 (changes requested):

cookbooks/fb_postfix/attributes/default.rb
jaymzh: Oops, the conditional above is supposed to be the other way around. Will fix. Nice catch.

PR #289, opened 2025-02-26
Enable MDL checking

jaymzh:
Had to change `.mdl.rules` - but should be compat with internal stuff.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,3 +57,11 @@ jobs:
         uses: actions/checkout@v4
       - name: Run Shellcheck
         uses: ludeeus/action-shellcheck@2.0.0
+  markdownlint:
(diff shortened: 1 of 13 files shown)


jaymzh:
Converting to draft while Meta internally sorts out the config files.

The style file was incorrectly opensourced as the rc file which won't work. That needs to be renamed, and the appropriate rc file opensourced. Once that's sorted this will work properly.

dafyddcrosby:
Should work after rebasing onto d1e9d6987ad0f43ab8e764e2915be9d2094c9b3c

jaymzh:
@dafyddcrosby - rebased, but I think your config isn't compatible with a newer version of MDL. I can update it to use the bundle version instead of the one from Actions, if you'd like.

PR #304, opened 2025-04-03
fb_iptables: minor cleanups

jaymzh:
* centos10 now has epel, don't skip packages
* Trailing whitespace

### DIFF
--- a/cookbooks/fb_spamassassin/recipes/default.rb
+++ b/cookbooks/fb_spamassassin/recipes/default.rb
@@ -123,13 +123,13 @@ end
 systemd_unit 'enable spam assassin update job' do
   only_if { node['fb_spamassassin']['enable_update_job'] }
   unit_name "#{update_job}.timer"
-  action [:enable]
+  action [:enable, :start]
 end
 
 systemd_unit 'disable spam assassin update job' do
   not_if { node['fb_spamassassin']['enable_update_job'] }
   unit_name "#{update_job}.timer"
-  action [:disable]
+  action [:stop, :disable]
 end
 
 service 'spamd' do


cookbooks/fb_iptables/recipes/packages.rb
```diff
@@ -19,14 +19,16 @@
...
+if node.centos_min_version?(9) || node.fedora?
   packages = ['iptables-legacy']
 else
   packages = ['iptables']
 end
+
 if node.ubuntu?
   packages << 'iptables-persistent'
+elsif node.centos_min_version?(9) || node.fedora?
+  packages << 'iptables-nft-services'
```
davide125, line 31: This doesn't seem right -- if we're using legacy on >= 9 we shouldn't use the nft services there. This cookbook probably needs an overall knob to do stock/nft vs legacy/not-nft. @michel-slm what do you think?

davide125 (dismissed):

cookbooks/fb_iptables/recipes/packages.rb
jaymzh: AFAICT, iptables-nft-servers *replaces* iptables-services. It literally `obsoletes` it, and on a fresh install of c9 or c10 you cannot install iptables-services, even if iptables-legacy is installed.

You can't even `dnf swap` them.

jaymzh: (also as it stands this doesn't support stock nft at all)

jaymzh: These files exist in ... both packages on both OSes.
```
/usr/libexec/initscripts/legacy-actions/ip6tables
/usr/libexec/initscripts/legacy-actions/ip6tables/panic
/usr/libexec/initscripts/legacy-actions/ip6tables/save
/usr/libexec/initscripts/legacy-actions/iptables
/usr/libexec/initscripts/legacy-actions/iptables/panic
/usr/libexec/initscripts/legacy-actions/iptables/save
```

Check this out - they're the same:

```
[root@scale-lists-centos9 ~]# rpm -qf /usr/libexec/initscripts/legacy-actions/iptables/panic
iptables-services-1.8.10-11.1.el9.noarch
iptables-nft-services-1.8.10-11.el9.noarch
[root@scale-lists-centos9 ~]# rpm -qV iptables-nft-services | grep legacy
[root@scale-lists-centos9 ~]# rpm -qV iptables-services | grep legacy
[root@scale-lists-centos9 ~]# 

```

michel-slm: note that Fedora does not even have nft-services https://src.fedoraproject.org/rpms/iptables/blob/rawhide/f/iptables.spec

jaymzh: @michel-slm - I think you're missing the point. I'm not supporting nft.

As it stands iptables-nft-services has *THE EXACT SAME FILES* in it as iptables-services, and if you install a C10 box with a kernel that supports good-old-iptables, and you try to install iptables-services **dnf will not let you**.

You **must** use `iptables-nft-services` with `iptables-legacy` to use LEGACY iptables.

michel-slm: ok, so per @jaymzh iptables-nft-services can actually manage the legacy backend too, so we should at least install that in c10s. Don't touch Fedora though, on Fedora the package is still called iptables-services (don't ask) - there is no iptables-nft-services

I'll help check tonight if the legacy backend even works on the stock kernel, I don't have my baremetal c10s host right now

michel-slm:
This looks correct now and the description seems fine, but let's hold on this until we get our internal systems' package management switched. We don't really want to end up installing the NFT backend's userspace on our systems which installing iptables-nft-services will do.

Also note that if we retire iptables-epel in Fedora EPEL `iptables-legacy` will be gone on CentOS Stream. so the question is - would this work for you in that case? e.g. can you use `iptables` with `iptables-nft-services` and keep your legacy rules? I think it will (with `CONFIG_NFT_COMPAT`) but worth testing

jaymzh:
On CentOS 10 it won't matter - the kernel doesn't support conntrack or other rules in xtables format.

On CentOS9, that would break things, so don't do that. :)

jaymzh:
OK I take it back, works fine in C10, even without legacy. PR updated.

cookbooks/fb_iptables/recipes/packages.rb
```diff
@@ -19,14 +19,20 @@
 # limitations under the License.
 #
 
-# c10s is opted out until EPEL 10 becomes available
-if (node.centos? && !(node.centos7? || node.centos8? || node.centos10?)) || node.fedora?
-  packages = ['iptables-legacy']
```
joshuamiller01, line 24: So now iptables-legacy is not required in any way - even on C9?

jaymzh: Correct. It's a strict subset of iptables-nft. And the overlapping files have the same checksum.

michel-slm: note that we need to switch our internal deployment away from this first - iptables-legacy is not needed if you're on the stock kernel (that has CONFIG_NFT_COMPAT enabled, so it can still delegate to the old XTABLES backend) but is needed still at Meta

michel-slm: also: on c9 and above you need iptables-nft, likewise on Fedora, but on EL8 - which is still supported (RHEL 8, Alma 8, Rocky, Oracle etc) - though idk if we want to support it - the package is still iptables

cookbooks/fb_iptables/recipes/packages.rb
```diff
@@ -19,14 +19,30 @@
...
+# In EL10/C10 and above, it's the same but iptables-nft
+# obsoletes iptables-services, so they cannot both be installed.
+if node.el_min_version?(9) || node.fedora?
+  packages = ['iptables-nft']
 else
   packages = ['iptables']
 end
-if node.ubuntu?
+
+if ::ChefUtils.debian?
```
davide125, line 34: This is breaking internally

```
/opt/chef/embedded/lib/ruby/gems/2.7.0/gems/chef-utils-16.18.30/lib/chef-utils/dsl/platform_family.rb:58:in `debian?': undefined method `[]' for nil:NilClass (NoMethodError)
```

davide125 (changes requested):

cookbooks/fb_iptables/recipes/packages.rb
jaymzh: Fixed.

jaymzh:
ShipIt didn't properly mark this as merged, but it was merged.

PR #310, opened 2025-04-25
fb_helpers cleanup

jaymzh:
I went to start adding missing version of all our osXX? version wrappers but then I thought - this isn't sustainable, we should come up with something better. We already have like centos10? but not rhel10? for example. We can do better.

So instead this if a new approach. For each platform we create a new _v method which will work for all versions. This is how it works:

* Create fedora_v?() which you can pass 39, 40, 41 to, etc.
* Turn fedoraXX? methods to just call the above, and mark them deprecated

And repeat for all platforms which we had these functions.

This also clears up rhel? vs rhel_family? (which were the same, but not made clear) and redhat?

It also fixes the tests which were setting node.default, but ohai attrs are node.automatic.

And final note: I didn't do windows helpers because their versions don't map directly, so there's no nice way to do it.

### DIFF
--- a/cookbooks/fb_helpers/README.md
+++ b/cookbooks/fb_helpers/README.md
@@ -39,23 +39,26 @@ your node.
 * `node.centos?`
     Is CentOS
 
+* `node.centos_version?(v)`
+    Is this CentOS version v
+
 * `node.centos5?`
-    Is CentOS Linux 5
+    Is CentOS Linux 5 - Deprecated: use `node.centos_version?`
 
 * `node.centos6?`
-    Is CentOS Linux 6
+    Is CentOS Linux 6 - Deprecated: use `node.centos_version?`
 
 * `node.centos7?`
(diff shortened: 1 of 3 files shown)


dafyddcrosby:
I like the approach, but at the risk of bikeshedding the tiniest bit, will say that `_v?` isn't descriptive, and instead should be `_version?`. This is something we're going to look at a lot and probably be making a Cookstyle cop to migrate existing code to. I can see `centos_v?(9)` becoming `centos?(9)` (or a similar typo) pretty easily, and want to avoid that future debugging session.

jaymzh:
Happy to do that, but, as I mentioned in Signal, I just wanna confirm that's the final name. I'm fine with whatever, but updating this PR is difficult because that file changes a lot and so there's a lot of manual merging that has to happen each time, so I"d like to make sure we're all on the same page, then do it when you'll have a moment to import+test+merge.

PR #311, opened 2025-04-25
Update resource_spec.rb

i5d6:
Use of Hardcoded Credentials

Access the code repository: If the repository is public or if they have unauthorized access to the codebase.

Extract the credentials: The attacker can see the hardcoded password, which can then be used to gain unauthorized access to the system.


i5d6 (commented):
:password => ENV['USER_PASSWORD'],

jaymzh (changes requested):
This is not a real password. No real users are being created. It's a unit test.

PR #331, opened 2025-07-24
Generate missing or out of sync metadata.json

skywalk7:
Summary:
X-link: https://github.com/facebook/IT-CPE/pull/283

Automatically generated metadata.json using `knife cookbook metadata --all`

Differential Revision: D75697334


skywalk7:
> You said automatic... is there some process that updates these automatically when metadata.rb is updated? I'm fine with that being an internal thing, but I'm just curious. If not, should we have some check that there's no delta on them?

Yes, we do have autocorrect lint tool. It's just initial push that's semi-auto. 

P.S. Sorry for the noise.

PR #362, opened 2026-04-01
Exclude spec from UseNodeDefault cop

dafyddcrosby:
Summary:
Spec files legitimately use node[] without node.default[], so exclude
them from the UseNodeDefault cop in the open-source cookstyle config. This addresses OSS CI
failures.

Differential Revision: D99126590

### DIFF
--- a/.cookstyle.yml
+++ b/.cookstyle.yml
@@ -554,6 +554,8 @@ Chef/Meta/TimeshardIsValid:
 
 Chef/Meta/UseNodeDefault:
   Enabled: true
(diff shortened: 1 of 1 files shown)


jaymzh (changes requested):
Sorry what am I missing here? why would a spec write to the node not through a writer?

More importantly, even if we did something like `node.override`, we probably want to disable/enable the rule around those (and they should likely be in `before` blocks), so that we get the value of these catching legit mistakes in specs.

PR #363, opened 2026-04-01
Fix rubocop:ignore to rubocop:disable

dafyddcrosby:
Summary:
rubocop:ignore is not valid rubocop syntax — the correct directive is
rubocop:disable. This fixes lint failures in fb_networkd, fb_storage,
and fb_systemd.

Differential Revision: D99135420


jaymzh (changes requested):
This is stacked on your other PR which I have concerns about, but this PR itself is fine.

We should consider an upstream Cop to catch `rubocopy:<not-a-directive>` errors, just like we catch disables on cops that don't exist.

PR #365, opened 2026-05-26
Migrate fb_apache to use chefutils helpers

jaymzh:
Built on top of #364 - move fb_apache to use the helpers there.


jaymzh:
tested on a few boxes with 0 delta.

PR #372, opened 2026-05-26
node.windows? migration in misc

jaymzh:
Last few places, all migrating in a single PR since they're minimal.


jaymzh:
Requires #373

PR #378, opened 2026-09-10
Apply JSON recipe substitutions on newer Chef versions

AadityaNair:
The newer versions have JSON recipes but not the substitutions.

This is a simple way to have this everywhere.


dafyddcrosby (changes requested):
I think we may want to narrow the version checks instead, so that we exclude the patches at 19.1.53/18.7.28 on everything but `from_hash` and `JSON_RECIPE_VALUE_SUBSTITUTIONS`. That way it's clearer to us in 2-3 years what's going on and our specific patches aren't yanked out inadvertently.

AadityaNair:
Addressed comments and updated

dafyddcrosby (commented):
This duplicates the logic below, and creates potential for inadvertent drift. Once that gets sorted, I think this should be in good shape 😃

dafyddcrosby (approved):

jaymzh (approved):

